PR #17, opened 2025-06-20
Adding Support to Google Gemini Models

marlonjsilva:
# Description

Adding support for Google Gemini models (`gemini-2.5-pro`, `gemini-2.5-flash`, `gemini-2.5-flash-lite-preview-06-17`) through their compatibility with the OpenAI API [link](https://ai.google.dev/gemini-api/docs/openai).

PR #30, opened 2025-08-03
Gemini support

lukaszkorecki:
Picks up from https://github.com/editor-code-assistant/eca/pull/17/files

## Authentication

- [ ] support for Gemini developer API 
- [ ] support for Gemini + Vertex AI (ADC & auth token)

## TBD

- [ ] make it work

### DIFF
--- a/src/eca/llm_api.clj
+++ b/src/eca/llm_api.clj
@@ -35,6 +35,7 @@
         (string/join "\n" (subvec lines start end)))
       content)))
 
+;; NTOE: this is not going to scale ;-)
 (defn ^:private anthropic-api-key [config]
   (or (:anthropicApiKey config)
       (config/get-env "ANTHROPIC_API_KEY")))
@@ -50,6 +51,25 @@
 (defn ^:private openai-api-url []
   (or (config/get-env "OPENAI_API_URL")
       llm-providers.openai/base-url))
+
+
+;; Google Gemini auth
+(defn ^:private gemini-api-key [config]
+  (or (:geminiApiKey config)
+      (config/get-env "GEMINI_API_KEY")))
+
+(defn ^:private google-api-key [config]
+  (or (:googleApiKey config)
+      (config/get-env "GOOGLE_API_KEY")))
+
+(defn ^:private google-project-id [config]
+  (or (:googleProjectId config)
+      (config/get-env "GOOGLE_PROJECT_ID")))
+
+(defn ^:private google-project-location [config]
+  (or (:googleProjectLocation config)
+      (config/get-env "GOOGLE_PROJECT_LOCATION")))
+
 
 (defn default-model
   "Returns the default LLM model checking this waterfall:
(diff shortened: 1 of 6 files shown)


src/eca/llm_api.clj
```diff
@@ -34,6 +35,7 @@
         (string/join "\n" (subvec lines start end)))
       content)))
 
+;; NTOE: this is not going to scale ;-)
```
ericdallo: yeah, I think we can easily create a util that receives the provider and do this check automagically, was waiting for more cases to understand usage

src/eca/llm_api.clj
```diff
@@ -168,6 +210,7 @@
             provider-fn (case (:api provider-config)
                           "openai" llm-providers.openai/completion!
                           "anthropic" llm-providers.anthropic/completion!
+                          "gemini" llm-providers.google/completion!
```
ericdallo: 👍🏻

src/eca/llm_api.clj
```diff
@@ -150,6 +171,27 @@
...
+        :instructions instructions
+        :user-messages user-messages
+        :max-output-tokens max-output-tokens
+        :reason-tokens reason-tokens
+        :reason? (and reason? (:reason? model-config))
+        :past-messages past-messages
+        :tools tools
+        :web-search web-search
+        ;; NOTE: no :api-url here, because Google provider figures it out
+        ;;       because it depends on how we're authenticated
```
ericdallo: that's new to me, interesting, I wonder if there is any case user want to customize the url, but it's not a problem right now

ericdallo (commented):
So far looking good, LMK if you wanna discuss any TODO, BTW we can start supporting basic chat features and leave MCP/reason and other features to next PRs what you think?

src/eca/llm_api.clj
lukaszkorecki: Yeah, I figured that until this point there was no need for more abstractions for providers but I had to leave that comment after adding all possible configs for Gemini - perhaps a provider protocol would make sense? It would also simplify the dispatch code - half of the args are needed only once (auth, URLs) and could be pre-computed on a record instance just once.

src/eca/llm_api.clj
lukaszkorecki: actually, I'll change that to Google - it will follow established pattern of using company names rather than products (OpenAI  vs ChatGPT or Anthropic vs Claude)

src/eca/llm_api.clj
lukaszkorecki: I did some research (well... Perplexity did most of it ;-)), and outside of running the code within GCP there's no real support for providing custom API urls, apart from custom telemetry endpoints. I'd say that we add a `TODO` and wait for someone to request it.

lukaszkorecki:
> So far looking good, LMK if you wanna discuss any TODO, BTW we can start supporting basic chat features and leave MCP/reason and other features to next PRs what you think?

I think that makes sense - with all of the plumbing out of the way, chat was the next thing that I was going to tackle

src/eca/llm_api.clj
```diff
@@ -80,19 +83,22 @@
...
+         (when-let [custom-provider-default-model (first (keep (fn [[model config]]
+                                                                 (when (and (:custom-provider? config)
+                                                                            (:default-model? config))
+                                                                   model))
+                                                               (:models db)))]
+           [:custom-provider-default-model custom-provider-default-model])
+         (when (anthropic-api-key config)
+           [:api-key-found "claude-sonnet-4-0"])
+         (when (openai-api-key config)
+           [:api-key-found "o4-mini"])
```
ericdallo: beaware this changed to gpt-5

src/eca/llm_api.clj
```diff
@@ -80,19 +83,22 @@
...
+         (when-let [ollama-model (first (filter #(string/starts-with? % config/ollama-model-prefix) (keys (:models db))))]
+           [:ollama-running ollama-model])
+         (when (google-any-auth? config)
```
ericdallo: maybe this check should come before ollama?

lukaszkorecki: yeah good call, I just updated the order - external APIs first, then ollama, then the default

ericdallo:
Closing this as outdated + we now have google support, thank you for the help so far!

PR #32, opened 2025-08-04
fix: document ProgressMessage

tomgeorge:
`ProgressMessage` was not documented in the protocol. It looks like there are only two possible states, `running` and `finished`.


ericdallo (approved):
Good catch, thank you!

Issue #62, opened 2025-08-19
Make context query case insensitve


afucher:
My experience in other tools is that when searching for context it is case insensitive, so when I want to add the `README.md` I can just type `readme`, instead of need to make it upper case

<img width="450" height="409" alt="Image" src="https://github.com/user-attachments/assets/bee4575b-479f-4ba5-9a37-403f1c41e378" />

<img width="438" height="408" alt="Image" src="https://github.com/user-attachments/assets/0f594304-19ae-4123-a962-6e2c6eca8b88" />
 

To upvote this issue, give it a thumbs up. See [this list](https://github.com/editor-code-assistant/eca/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

Issue #61, opened 2025-08-19
Allow add file (read-only) outside of workspace


jilen:
Useful for add file from git submodule, or other library source file.

ericdallo:
Interesting, makes sense, I wonder how we should present those files in context completion, any suggestions?

ericdallo:
I'm finding hard how the UX would be for this one, we can't just return all system files for completion and there is no option for manual add contexts not in completion.

It seems Cursor has the same problem, so the workaround for now is to add the folder as a workspace to ECA be aware of it.

jilen:
Maybe just return child items in the folder if prefix is `/`,  `./` , `../`, `~/` ...

ericdallo:
hum, maybe, will take a look if with that would work

r0man:
Some ideas for a Emacs user interface:
- Use find-file/project-find-file or a custom function
- Add files from dired

For other editors
- Use some file picker where you can select 1 or many files

ericdallo:
I think the problem here is that:
- clients request server for what available contexts exists and they should be able to complete/choose via `chat/queryContext`
- server return the contexts checking for workspace root.

otherwise we may have too much login on clients and not centralized, so would be nice to fix that somehow in the query, like user type `~/` or `../` and server realizes that and return that folder in completion

ericdallo:
Ok, I think I have something work for this in server with minor changes in client and seems to fix the problem

ericdallo:
Done! LMK if that is enough for you @jilen

PR #66, opened 2025-08-19
change query context to be case insensitive

afucher:
Fixes https://github.com/editor-code-assistant/eca/issues/62

Change how the query filter of contexts is made to make it case insensitive

make glob generic, getting all files
filter manually using includes? and lower-case
Add tests for context

🤖 Generated with [eca](https://eca.dev/)


ericdallo:
Awesome, I believe this will help unblock #61

ericdallo (approved):
looks great!

Issue #63, opened 2025-08-19
Make rules files agnostic of extension


ericdallo:
To upvote this issue, give it a thumbs up. See [this list](https://github.com/editor-code-assistant/eca/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

we don't need for check for specific file extensions, we should probably recommend only .md though

bombaywalla:
Would you say something more about this issue?  Which file(s) should I look at?  I looked at `rules.clj` and I don't see any mention of a file extension.  Thanks.

ericdallo:
Hey @bombaywalla good catch, actually the code is already agnostic of extension, I believe we can just update docs to not mention `mdc`, only `md` as recommendation maybe

bombaywalla:
Thanks.  I believe `.mdc` is just a `.md` file but with meta-data at the beginning.
Cursor, apparently, uses `.mdc` files for its rules.
So, if we're recommending anything, then perhaps `.md` and/or `.mdc`?

ericdallo:
I'm not sure what .mdc stands for, I heard it could be `markdown cursor`, so I don't want to couple with it.
I believe the file name is less of the problem, for this issue we can just update the docs.

But you may be interested in #26 as we are right now ignoring those metadata, not parsing or using with LLM

bombaywalla:
My understanding also is that `.mdc` stands for `markdown cursor`.
Even if we currently ignore the metadata, I think that supporting, and perhaps even recommending, `.mdc` files as well as `.md`files is a good idea, so that developers can use the same files with Cursor and with ECA.

Happy to make whatever changes you decide on.

ericdallo:
I don't like to recommend a file extension couple to another tool, it would be best to suggest `.md` and support `.mdc` if available (which we already do), WDYT?
So, not mention .mdc in docs is already enough to me

bombaywalla:
It seems that `mdc` is only mentioned in `configuration.md`.
Is that your understanding as well?

Since the parsing of a `.mdc` file seems to be exactly the same as the parsing of a `.md` file (please correct me if I am wrong), we could just remove any reference to `mdc` and adjust the rest, if needed.  Which seems like what you were saying in the first place.  If you agree with that approach, I'll go and make a PR based on it.

ericdallo:
yes!

exactly, go ahead!

bombaywalla:
Please review PR #72

[...]

PR #72, opened 2025-08-22
Rule file extension docs

bombaywalla:
Addresses #63.

The code accepts rule files with any extension.
This PR updates the docs to reflect that.
It also removes mention of `.mdc` files as the preferred format.


ericdallo (approved):
Thank you!

PR #74, opened 2025-08-23
Fix hanging LSP diagnostics requests with configurable timeout

zikajk:
Prevents ECA from hanging indefinitely when editor diagnostic requests don't respond by adding a configurable timeout mechanism.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -18,7 +18,7 @@ There are multiples ways to configure ECA:
       }
     }
     ```
-  
+
 === "Local Config file"
 
     Convenient for users
@@ -31,7 +31,7 @@ There are multiples ways to configure ECA:
       }
     }
     ```
-    
+
 === "InitializationOptions"
 
     Convenient for editors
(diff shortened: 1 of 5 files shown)


docs/configuration.md
```diff
@@ -232,6 +233,7 @@ There are 3 possible ways to configure rules following this order of priority:
         "manualApproval": null,
       },
       "mcpTimeoutSeconds" : 60,
+      "diagnosticTimeoutSeconds" : 30,
```
ericdallo: I wonder if this timeout should be something more general for requests to editor on LLM call, which we will have more I believe for references, definition and others, WDYT?
Maybe `EditorTooCallTimeoutSeconds`,
I thought about `toolCallTimeoutSeconds` if we manage to make it generic for other tool calls but not sure is a good idea.

zikajk: Do you think we'll have more configuration options for LSP in the future? If so, I would make it `{"lsp": {"timeout": 30}}` like an Opencode. But if not, what about just changing it to `lspTimeout`?

ericdallo: Yeah, I 'm ok with something generic for lsp for now like `lspTimeout`, let's go with that

zikajk: Updated 👍

ericdallo:
Thank you!

PR #80, opened 2025-08-24
Support tool call details for clojure-mcp

### LINKED ISSUE #67: Support ECA tool diffs for other MCP tools

Like for clojureMCP which has file write and other tools similar.
We should support in a way that user can configure and maybe eca can have good defaults.

r0man:
Fixes #67 

This changes the `get-tool-call-details` into a multi method that can be specialized by tool name. It also adds a `get-tool-call-details-result`.

The `get-tool-call-details` method is called BEFORE the tool is called and the tool gets the chance to add information to inform the user that the tool is about to be called.

The `get-tool-call-details-result` method is called AFTER the tool is called and the tool gets the chance to add aditional information to the details from the `get-tool-call-details` and the actual result.

I tested this with the following prompts, and I do now see pretty diffs in the chat buffer.

```
create the test.clj file with the content "{:version "1.0.0"}"
bump the version in test.clj. you must use the file_edit MCP tool
bump the version in test.clj. you must use the file_write MCP tool
bump the version in test.clj. you must use the clojure_edit_replace_sexp MCP tool
bump the version in test.clj. you must use the clojure_edit MCP too
```

This PR needs the following change in Emacs (and possibly other clients):

https://github.com/editor-code-assistant/eca-emacs/pull/19

In my Emacs it looks like this:
image

### DIFF
--- a/src/eca/diff.clj
+++ b/src/eca/diff.clj
@@ -55,3 +55,28 @@
       (->> (DiffUtils/generateUnifiedDiff file file original-lines patch 3)
            (drop 2) ;; removes file header
            unlines)})))
+
+(defn unified-diff-counts
+  "Given a unified diff string, return a map with counts of added and removed lines.
+
+  It ignores diff headers (---, +++), hunk markers (@@), and metadata lines starting with \\."
+  [diff-text]
+  (let [lines (string/split-lines diff-text)]
+    (reduce (fn [{:keys [added removed] :as acc} line]
+              (cond
+                (or (string/starts-with? line "---")
+                    (string/starts-with? line "+++")
+                    (string/starts-with? line "@@")
+                    (string/starts-with? line "\\")
+                    (string/blank? line))
+                acc
+
+                (string/starts-with? line "+")
+                (update acc :added inc)
+
+                (string/starts-with? line "-")
+                (update acc :removed inc)
+
+                :else acc))
+            {:added 0 :removed 0}
+            lines)))
--- a/src/eca/features/chat.clj
+++ b/src/eca/features/chat.clj
@@ -188,7 +188,7 @@
                          (let [calls (doall
                                       (for [{:keys [id name arguments] :as tool-call} tool-calls]
                                         (let [approved?* (promise)
-                                              details (f.tools/get-tool-call-details name arguments)
+                                              details (f.tools/tool-call-details-before-invocation name arguments)
                                               summary (f.tools/tool-call-summary all-tools name arguments)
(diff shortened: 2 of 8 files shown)


r0man:
This seems to work, but:
- I'm not sure if we should use 1 or 2 multi methods here (we could add an argument indicating before/after)
- maybe rename the functions to a better name that makes the before/after tool call more clear? 
- I'm using the "fileChange" detail structure which may indicate that a file is about to change, maybe we should add a "fileChanged" detail structure that indicated that a file was changed
- I see that ECA prefixes its tool names, clojure mcp does not. this might lead to conflicts when specializing multi methods. maybe we can talk to @bhauman if it would make sense to add a clojure_mcp_ prefix to the tools.

src/eca/diff.clj
```diff
@@ -55,3 +55,28 @@
       (->> (DiffUtils/generateUnifiedDiff file file original-lines patch 3)
            (drop 2) ;; removes file header
            unlines)})))
+
+(defn unified-diff-counts
+  "Given a unified diff string, return a map with counts of added and removed lines.
+
+  It ignores diff headers (---, +++), hunk markers (@@), and metadata lines starting with \\."
```
ericdallo, line 62: good

bhauman:
The prefixing is a detail left to clients. Claude Code and Claude Desktop both prefix the tools with clojure-mcp although Claude Desktop hides the prefix in the interface.

ericdallo:
Thank you a lot @r0man for this issue!

> maybe rename the functions to a better name that makes the before/after tool call more clear?

I like that, have a different function for before and another for later

> I'm using the "fileChange" detail structure which may indicate that a file is about to change, maybe we should add a "fileChanged" detail structure that indicated that a file was changed

Hum, yeah not sure, I'd avoid this change until really needed

> I'm using the "fileChange" detail structure which may indicate that a file is about to change, maybe we should add a "fileChanged" detail structure that indicated that a file was changed

Thanks @bhauman! I think we should create a issue for prefixing tools @r0man, sounds like a good idea indeed
For now we can go with no prefix in this PR, or somehow receive the tool server too .

ericdallo (approved):
Looks great now, will test and then merge, we should publish on clojurians #ai-assisted-tools that ECA is the first LLM client with better UX support for clojureMCP

ericdallo:
@r0man did some testing and realized that maybe we should support `clojure_edit` as well? We can do in another PR if you prefer

ericdallo:
Nvm, I didn't update eca-emacs, works great!

PR #88, opened 2025-08-27
Fix minor grammatical errors in eca_base.txt

ArthurHeymans:


ericdallo:
@zikajk FYI if you are touching that code

Issue #83, opened 2025-08-25
Improve toolCall approval behavior


afucher:
Today ECA default for toolCall is to allow all by default, my previous experience with other similar tools is to have the default as manual approve (Claude, Cursor).
This is a huge difference and I was not expecting this at all, it just changed a lot of files, called tools, and more, without I accept.

My suggestion is to:
 - Change the default behavior to be manual Approval;
 - On the client side, have along with `Reject` and `Run` buttons, something similar to what we have in other tools to allow the specific action always, so is easy to configure while you are using it.

Another option that Claude has is to add a new mode to plan/agent that is a mode that allows all tools.

For new users to Code Assistants or users that are migrating to ECA, I think that this proposed experience will be better.

To upvote this issue, give it a thumbs up. See [this list](https://github.com/editor-code-assistant/eca/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

ericdallo:
I don't have a strong opinion on manualApprove, so I'd love to hear from @editor-code-assistant/maintainers .

The default approval is easy to change, but create another button or communication to always approve that tool is not, it requires changes in protocol, although not hard

r0man:
Nothing bad has happened (yet) to me, but I also think we should have a safer default. 

I would do the same as Claude Code does. Ask the user for approval, with the option to remember previous decisions. In one of the videos I watched about Claude Code they say, this was actually one of the more involved features. I think they also ask the LLM (a fast model, like Haiku) if a command is safe to run.

Additionally I would like to give ECA access to some other directory in which it can do operations. With Claude Code I sometimes check out a repo and tell it to look there for additional information.

ericdallo:
@r0man you may be interested in #61

zikajk:
I'd just like to add that, for me, it's not about adding files to the context. I'd like to be able to use e.g., a bash tool to operate within a different directory that I point to in the prompt, which is currently not possible. 
Of course, it would be good to have an approval step for this to prevent dangerous operations.

r0man:
I'm also more interested in what @zikajk said. Running a bash tools in an other directory, not adding files to the context. ECA should be bale to gather information from say a checkout of an other related project. If I write a prompt like:
`Take a look at the files in directory /home/roman/workspace/other-project`
ECA should ask if it is ok to run grep/bash etc in that dir. Like Claude Code does.

ericdallo:
[...]

PR #95, opened 2025-08-29
Add toolCall approval

ericdallo:

### DIFF
--- a/src/eca/config.clj
+++ b/src/eca/config.clj
@@ -57,6 +57,16 @@
                          :excludeCommands []}
                  :editor {:enabled true}}
    :disabledTools []
+   :toolCall {:approval {:allow {"eca_shell_command" {}
+                                 "eca_editor_diagnostics" {}
+                                 "eca_directory_tree" {}
+                                 "eca_write_file" {}
+                                 "eca_edit_file" {}
+                                 "eca_preview_file_change" {}
+                                 "eca_move_file" {}
+                                 "eca_read_file" {}
+                                 "eca_grep" {}}
+                         :ask []}}
    :mcpTimeoutSeconds 60
    :lspTimeoutSeconds 30
    :mcpServers {}
@@ -160,7 +170,9 @@
                  (fn [providers]
                    (when providers
                      (-> (normalize-providers providers)
-                         (update-vals normalize-provider-models)))))))
+                         (update-vals normalize-provider-models)))))
+      (update-in [:toolCall :approval :allow] update-keys name)
+      (update-in [:toolCall :approval :ask] update-keys name)))
 
 (defn all [db]
   (let [initialization-config @initialization-config*
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -187,7 +187,10 @@ There are 3 possible ways to configure rules following this order of priority:
         };
         disabledTools: string[],
         toolCall?: {
-          manualApproval?: boolean | string[], // manual approve all tools or the specified tools
+          approval: {
+            allow: {{key: string}: {}},
+            ask: {{key: string}: {}},
+          };
         };
         mcpTimeoutSeconds: number;
         lspTimeoutSeconds: number;
@@ -234,7 +237,13 @@ There are 3 possible ways to configure rules following this order of priority:
                        "editor": {"enabled": true}},
       "disabledTools": [],
       "toolCall": {
-        "manualApproval": null,
+        "approval": {
+          "allow": {
+            "eca_edit_file": {}
+            // ... all eca tools
+          },
+          "ask": {},
+        }
       },
       "mcpTimeoutSeconds" : 60,
       "lspTimeoutSeconds" : 30,
--- a/src/eca/features/tools.clj
+++ b/src/eca/features/tools.clj
@@ -117,15 +117,32 @@
                                                                         (assoc :type :mcp)
                                                                         (update :tools #(mapv with-tool-status %)))))})))
 
+(defn legacy-manual-approval? [config]
+  (let [manual-approval? (get-in config [:toolCall :manualApproval] nil)]
+    (if (coll? manual-approval?)
+      (some #(= name (str %)) manual-approval?)
+      manual-approval?)))
+
 (defn manual-approval? [all-tools name args db config]
   (boolean
-   (let [require-approval-fn (:require-approval-fn (first (filter #(= name (:name %))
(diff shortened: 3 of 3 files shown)


src/eca/config.clj
```diff
@@ -57,6 +57,17 @@
...
+   :toolCall {:approval {:allow {"eca_shell_command" {}
+                                 "eca_editor_diagnostics" {}
+                                 "eca_directory_tree" {}
+                                 "eca_write_file" {}
+                                 "eca_edit_file" {}
+                                 "eca_preview_file_change" {}
+                                 "eca_move_file" {}
+                                 "eca_read_file" {}
+                                 "eca_grep" {}}
+                         :ask {"eca_shell_command" {:argsMatchers {"command" [".*rm\\s.*"
```
ericdallo: Didn't spent too much time on these defaults, suggestions for better default are welcome, we can change later too

src/eca/config.clj
```diff
@@ -127,47 +138,67 @@
 
 (def ollama-model-prefix "ollama/")
 
-(defn ^:private normalize-providers [providers]
```
ericdallo, line 130: Not related with PR but required this refactor to fix naming in configs

src/eca/config.clj
zikajk: Do you think we need the `"command"` part? I'm wondering if `"patterns"` would be enough instead whole of `{:argsMatchers {"command" .... }}`.

I also wonder if it makes sense to have both `:ask` and `:allow`. 
If I have `{:allow {"eca_shell_command" {:argsMatchers {"command" [".*cp\\s.*"]}}}}` and also `{:ask {"eca_shell_command" {:argsMatches {"command" [".*rm\\s.*"]}}}}`, what happens with all the other commands? Or when I don't have `eca_grep` in any of them? Maybe I'd do it so we only have `allow` and everything else is `ask` - default? So, if allow doesn't contain a pattern, all `eca_shell_commands` would be allowed; if it has a pattern, only that one would be automatically allowed. WDYT? 

_I also think that the project settings should override the system settings (and then custom-behavior would override the global ones). With a single key, it will be simpler, and you won't have to think about what the consequence would be if you have _:ask_ at the system level but don't use it in the project._

And maybe I wouldn't remove the option to have all tools automatically allowed/denied, at least until we have some "approve and save" system. At the same time, I can imagine having a "dangerous" custom behavior that allows absolutely everything. That option would "overwrite" anything you have inside `:approval`. This would be useful, for example, if you want to run a well-defined plan.

zikajk: 1) Do you think the `"command"` part is necessary? I'm wondering if `"patterns"` would be enough instead of the entire `{:argsMatchers {"command" .... }}` structure.
I like the simple approach here -> https://opencode.ai/docs/permissions/

   _DOC note: I probably prefer the current reversed structure of `approval-type -> command` over `command -> approval-type`, but with that approach, users will have to understand the precedence rules. With a simple list, it's more straightforward because the last item takes precedence._ 

2)  I wouldn't remove the option to automatically allow all tools, e.g.`{:toolCall {:autoApprove? true}`. 
I can imagine a "dangerous" custom behavior that allows everything. This would be useful, for instance, when executing a well-defined plan.
I believe that this is where `:ask` shines - when everything is allowed by `autoApprove`, `:ask` can be used for dangerous things like `git push`, etc.
But of course, `:ask` is likely useful in other scenarios as well. For example, if I use `:allow {"eca_shell_command" {}}` to allow every shell command, I might still want one or two specific commands to not be automatic.

3) I think that It makes sense to add `:deny` option as well. Such a command should not be executed, and there shouldn't even be a prompt to run it. This way, even with `:autoApproval` set to `true`, these commands would still be rejected. My suggestion is that `:deny` should override `:ask` and `:ask` should override `:allow`. 
WDYT? 
**Edit:** Sorry, I just noticed that you already mentioned `deny` in your todo list 😸 .

ericdallo: 1. Hum, the problem is how to make it customizable for MCPs and other tools with custom args? Like if we want `clojure_bash` tool to only specific args? Also, how to customize other args, for example the `working_directory` of the `eca_custom_command` in case in specific dirs we want to ask?
If you can see that working, can you show a  an example config which would work ?

2. The point is we don't need it, if there is no config in :allow :ask or :deny, we default to approve the tool, it's already the current behavior, does that works? and currently it's like that, it's a waterfall, checking for ask first if not found on allow, if not found default to approve, maybe we don't even need the :allow? 

3. yep, hehe

zikajk: We've touched on this before, but I believe the [default behavior should be to ask](https://github.com/editor-code-assistant/eca/issues/83) for approval for everything. This would prevent them from having to manually list every single tool or MCP server. That said, if a user wants to bypass this, they should be able to set autoApprove = true.

Personally, I feel that a more detailed configuration is overkill. For the Bash tool, it makes sense to specify commands, but for all other tools, it should be enough to simply allow the tool itself without further restrictions. I'm not sure what you mean by `eca_custom_command`.
However, I'm not against this level of flexibility if it's beneficial to others. If I can leave the configuration map empty and it defaults to the tool with any arguments, then that seems like a good compromise.

For context, here is my local allow configuration for CC:
```
[
  "mcp__clojure-mcp__read_file",
  "mcp__clojure-mcp__fs_grep",
  "Bash(grep:*)",
  "mcp__sequential-thinking__sequentialthinking",
  "mcp__clojure-mcp__glob_files",
  "mcp__clojure-mcp__grep",
  "Bash(rg:*)",
  "mcp__clojure-mcp__scratch_pad",
  "Bash(clj-kondo:*)"
]

```

zikajk: How does it currently work with MCP? Let's say I have an MCP with thirty tools. If I wanted each of them to require approval (assuming they are approved by default), would I just need to use the server name and not have to specify each tool? But i am still afraid that it is easy to forget about it which can be costly :-).

ericdallo: eca_custom_command == eca_shell_command 😅 my bad

ericdallo: Owe discussed and came in agreement with the one I pushed, example:
```
    "toolCall": {
        "approval": {
            "byDefault": "allow",
            "ask": {"eca_shell_command": {"argsMatchers": {"command": [".*rm.*"]}},
                    "eca_edit_file": {},
                    "clojure-mcp__edit_file": {}}
        }
    },
```

Issue #92, opened 2025-08-27
Support user custom tools


CsBigDataHub:
When using GPTEL, I prefer using emacs native tools instead of MCP exposed tools.

In this instance I use this emacs tool to fetch content from URL rather than using [mcp-fetch](https://github.com/ExactDoug/mcp-fetch).

```emacs-lisp

  (gptel-make-tool
   :function (lambda (url)
               (with-current-buffer (url-retrieve-synchronously url)
                 (goto-char (point-min))
                 (forward-paragraph)
                 (let ((dom (libxml-parse-html-region (point) (point-max))))
                   (run-at-time 0 nil #'kill-buffer (current-buffer))
                   (with-temp-buffer
                     (shr-insert-document dom)
                     (buffer-substring-no-properties (point-min) (point-max))))))
   :name "read_url"
   :description "Fetch and read the contents of a URL"
   :args (list '(:name "url"
                 :type string
                 :description "The URL to read"))
   :category "web")

```

Now LLM uses `read_url` tool when I provide a URL.

ericdallo:
Interesting use case, I believe supporting user specify tools by themselves would be nice, but not sure it should be something done in client, but on server, and with that you wouldn't be able to do via elisp I guess.

For this case especifically we will implement https://github.com/editor-code-assistant/eca/issues/70.

ericdallo:
Any guess @editor-code-assistant/maintainers on how to approach this feature support?

zikajk:
I've thought about this. While I like that feature of GPTEL, it doesn't really fit into the overall design of ECA in my opinion. It would be limited to Emacs. If a user needs custom tools, they can write them in Clojure and recompile ECA, or they can write a Bash script and call it using the Bash tool.

ericdallo:
Yeah, agree, or even have MCPs for that but I agree is not a sconvenient than write a simple elisp.

Maybe we could have a way to specify custom tools scripts, that would reduce the work for users plug custom script as tools?

```javascript
{
"extraTools": {
   "websearch": {"bash": "/path/to/my-script"}
}
}
```

ChetanKoneru:
Makes sense, build tool in the server and use it in any client.

Lets say I want to create a tool to fetch url content in markdown with [Trafilatura](https://github.com/adbar/trafilatura), how would I achieve that.

bash script? still wouldn't we want to pass url as a parameter?

Some example scenarios in the documentation would be nice. 

[...]

Issue #79, opened 2025-08-24
Introduce a behaviour Configuration Key


zikajk:
To upvote this issue, give it a thumbs up. See [this list](https://github.com/editor-code-assistant/eca/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

1) I'm proposing a new top-level configuration key, `behaviour`, to resolve conflicts between the system prompt and a behavior. This change would organize the agent's configuration, making it more modular and easier for users to create custom behaviors as well.

The proposed structure would look like this:
```
{
  "behaviour": {
    "agent": {
      "promptFile": "path/to/prompt.txt" // if no `promptFile` for builtin provider is provided, use the default one
    },
    "plan": {...}
  }
}
```
2) Move most of the instructions from eca_base.txt to agent_behavior.txt and remove any configuration related to one common system prompt.

**Motivation**:
While creating the prompt for the plan behavior, I noticed it conflicted with some instructions in the system prompt. As I started moving those elements into the agent behavior prompt, there was very little left in the original system prompt, leading me to believe a file for the main system prompt is now unnecessary.

**RELATED FEATURES**:
I believe that adding some configuration options (e.g. defaultModel, ...) might solve those:
https://github.com/editor-code-assistant/eca/issues/58
https://github.com/editor-code-assistant/eca/issues/52

Simillar to this implementation - https://opencode.ai/docs/modes/

zikajk:
@ericdallo WDYT?

Do you agree that there is no reason to keep a single system prompt with the {behaviour} placeholder?
1) I'm afraid the instruction for users would be confusing if we keep it: e.g. "If you have a custom system prompt, do not forget at {behaviour} placeholder otherwise behavior modes won't work."

2) I've also noticed that markdown and tool calls work fine even when I don't use eca_base.txt.

ericdallo:
@zikajk yeah that config sounds good.
About eca_base, I'm ok if we keep with good default prompts, like the eca I traduction, how to communicate etc. if we intend to solve that completely moving to behaviors, I think it's ok

ericdallo:
Done!

PR #102, opened 2025-08-30
Support user custom tools -	https://github.com/editor-code-assistant/eca/issues/92

CsBigDataHub:


ChetanKoneru:
@ericdallo , please proofread, first time coding in Clojure, used Claude to improve code and generate unit-test.

docs/configuration.md
```diff
@@ -152,6 +152,52 @@ There are 3 possible ways to configure rules following this order of priority:
...
+          "name": {
+            "type": "string",
+            "description": "Name to greet"
+          }
+        },
+        "required": ["name"]
+      },
+      "description": "Say hello to someone",
+      "requireApproval": false,
+      "enabledInPlanMode": true
```
ericdallo, line 179: We should remove these.
`requireApproval`:. we have #95 which is close to be merged and will improve tools permissions config, this one should be handled there OOTB.
`enabledInPlanMode`: we should handle those in #79, which we will be able to have custom behavios and decide which tools are available in which ones.

docs/configuration.md
```diff
@@ -152,6 +152,52 @@ There are 3 possible ways to configure rules following this order of priority:
...
+- `requireApproval`: (optional) If true, requires user approval before running.
+- `enabledInPlanMode`: (optional) If true, tool can be used in plan mode.
+
+### Security
+
+User tools run as local shell scripts. Only add tools you trust. Use `requireApproval` for sensitive commands.
+
+### Usage
+
+Once configured, user tools are available like built-in tools. You can invoke them by name (e.g., `/tool hello name=World`).
```
ericdallo, line 197: This is weird, we should not support a `/tool` at all, tools are invoked automatically by LLMs and we pass them during the calls, check `eca.features.tools/all-tools`, tools __should not__ be called manually, sounds like LLM halucination

src/eca/features/tools.clj
```diff
@@ -19,38 +19,75 @@
...
+(defn ^:private substitute-workspace-roots [db description]
+  (string/replace description #"\\$workspaceRoots"
+                  (constantly (tools.util/workspace-roots-strs db))))
+
+(defn ^:private transform-tool-definition [db [name tool]]
+  [name (-> tool
+            (assoc :name name)
+            (update :description (partial substitute-workspace-roots db)))])
+
 (defn ^:private native-definitions [db config]
```
ericdallo, line 55: we should keep this function as it is, and add a new one for the custom tools in [this concat](https://github.com/editor-code-assistant/eca/pull/102/files#diff-60301e9fa63d939355dc6a869677d9d75758758773918ba192f17888c9aef79cR91)

ChetanKoneru: Thank you will do the changes

docs/configuration.md
zikajk: Yeah, I agree. 
You can always create a command with a prompt like "Call a ToolABCD with ..." if you need to have a shortcut.

src/eca/features/tools.clj
```diff
@@ -19,38 +19,75 @@
 
 (def ^:private logger-tag "[TOOLS]")
 
+;; ============================================================================
```
ericdallo, line 22: please let's avoid changing a lot of stuff here:
- those comment headers are unecessary
- focus on adding the tool in the concat like I mentioned

src/eca/features/tools/user.clj
```diff
@@ -0,0 +1,169 @@
+(ns eca.features.tools.user
```
ericdallo, line 1: let's call `eca.features.tools.custom` instead of user.

test/eca/user_test.clj
```diff
@@ -0,0 +1,153 @@
+(ns eca.features.tools.user-test
```
ericdallo, line 1: the ns doesn't match the file name

ericdallo:
@CsBigDataHub the idea in config is valid but the code is bad (because of LLM generated):

- we need tests fixed
- check my comments, we should do 2 major changes for enabling this issue:
  - make available custom tools in the concat I mentioned, geting from another `eca.features.tools.custom` ns.
  - when LLM call a tool in the `call-tool!` function, understand that is a custom tool and delegate to `eca.features.tools.custom/call-tool!`
- try creating unit tests following other tests in the project.

src/eca/features/tools/user.clj
```diff
@@ -0,0 +1,169 @@
...
+
+(set! *warn-on-reflection* true)
+
+(def ^:private logger-tag "[TOOLS-USER]")
+
+;; ============================================================================
+;; Schema Validation
+;; ============================================================================
+
+(s/def ::arg-type #{"string" "number" "boolean" "array" "object"})
```
ericdallo, line 17: Can we do not any schema validation here? I think we should:
- do schema validation only on `/doctor`, but please, let's do this in another PR, as we want more validations like config json etc.

ChetanKoneru:
@ericdallo yes, I agree, I will close this and give it crack after I practice more.

I will close this PR.

Issue #65, opened 2025-08-19
[bug] Stop prompt does not change the status of the last running message


afucher:
If ECA is running a command and I click to stop, the icon stay the loading one.

<img width="491" height="130" alt="Image" src="https://github.com/user-attachments/assets/87b86c0a-25e3-4517-8b35-02081049db49" />

<img width="482" height="148" alt="Image" src="https://github.com/user-attachments/assets/3524a53a-e647-4afb-93da-192d27d3f6bb" />

Suggestion:
Change the icon to something that indicate that it is stopped.

ericdallo:
Yeah, Good catch, I think we should fix this in server to provide a `toolCallRejected` when stop is pressed, this would fix for all editors too

bombaywalla:
Would you please provide more context on this issue?

From what I understood you propose that a `toolCallReject` message be sent by the client to the server when `stop` is pressed by the user.  That makes sense.  But then I'd expect those changes to be made on each editor client, and not on the server.  What am I missing?

bombaywalla:
From what I understand, when the user presses the `stop` button, the editor sends a `toolCallReject` to the server.
What you're suggesting is that on receiving a `toolCallReject`, the server then sends a `toolCallRejected` to the editor.
And then the editor can change the icon as required.

Assuming my understanding is correct, then in `features/chat.clj` it appears to me that the code is already implementing that desire.

Is that correct?  Or am I missing something?

ericdallo:
@bombaywalla, you are absolutely correct on how it works, the problem is when user stop while no toolCallRun was emmited, so we stay in a loading state

bombaywalla:
Okay.  So, you're suggesting that on getting `toolCallReject`, the server calls `toolCallRejected` whether or not `toolCallRun` was sent.  Is that correct?

ericdallo:
No the problem is not with `toolCallReject`, but with the `chat/promptStop` message that client sends when clicking on stop, when we receive that we should send a toolCallRejected if a toolCall is in progress

The stop OP mentions is not stop the tool call, but stop the chat prompt entirely

bombaywalla:
Ah, I missed that.
I'll take a closer look at the code.
Thanks.

bombaywalla:
[...]

PR #104, opened 2025-08-31
Handle promptstop and toolcall

bombaywalla:
This is a work-in-progress of the changes required for issue #65
Working on tests.

- [ ]


src/eca/features/chat.clj
```diff
@@ -42,12 +42,238 @@
...
+  Active tool calls are those not in the following states: :completed, :rejected, :stopped."
+  [db chat-id]
+  (->> (get-in db [:chats chat-id :tool-calls] {})
+       (remove (fn [[_ state]]
+                 (#{:completed :rejected :stopped} (:status state))))
+       (into {})))
+
+;;; Event-driven state machine for tool calls
+
+(def ^:private tool-call-state-machine
```
ericdallo, line 69: I wonder if this state machine is really needed / if it's not adding too much complexity, I got the point, but not sure it's the ideal

bombaywalla: I had similar thoughts while I was developing this patch, so I understand where you're coming from.  In the end, after having tried a different approach which did not use a state machine,  I felt that this approach is better because it affords better reasoning and extensibility - at a slight cost in the additional baggage of a framework.  I would not say that this approach is adding complexity - rather the opposite.

My thinking is as follows.  It seems to me that some kind of state management is needed.  Previously, we had several variables that maintained state of what notifications were sent, among other things.  Initially I tried to extend that, but the control logic got much more hairier and it was difficult to reason about.  Now there is only one state variable and a lot, but not all, of the control logic is formalized into a state transition machine that is specified by pure data.  I find it a lot easier to reason about now.  The code size is smaller.  It is also easier to extend - for example, to the case where we want to be able to interrupt a long-running call.  I found that most of the non-inherent complexity that is in this state machine is because I was trying to maintain the structure of the current code - in an attempt at increasing my confidence that I was not impacting the existing semantics for the rest of the system.  For example, the use of promises and futures.

I think the fundamental issue is that there is currently very little semantics documented about the protocol.  If such a semantics were defined (say, by using network sequence diagrams, as in the documentation of the Agent Client Protocol), then the implementation would naturally follow that diagram.  One of the benefits of this current approach is that it provides such a "specification" - albeit in code.

What do you think?

ericdallo: Thank you for the explanation @bombaywalla, yeah, I think the state machine makes sense in the end on a second thought, I will review when finished with that in mind!

About the protocol, I pretty much follow the standard of [LSP protocol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/) which is famous and used widely, but if we think we can improve even better not poluting too much the protocol, I'd accept thankfully.

bombaywalla: Thanks for responding.  I would love to have your review, but I should caution that it is not quite ready.  For example, I am somewhat surprised that integration tests are failing.  Do you have any insight as to why they are failing?  I took extra care, as you can see from the diffs, to ensure that I was not changing existing semantics that should not change. I also am in the process of adding new unit tests for the state machine.

I think I was not clear in communicating what I find lacking in the protocol specification.  I see that you are mimicking the LSP spec.  My biggest concern is that the current ECA spec does not specify clearly what valid sequences of messages are, and which sequences are invalid.  Which messages can come before/after other messages, for example. I think it would make validation testing of different implementations of the ECA protocol easier and more defined.  I might take a look at that after I finish with this PR.

ericdallo: > My biggest concern is that the current ECA spec does not specify clearly what valid sequences of messages are

I see, yeah, I think we should mention that somewhere for sure, the ones you found about the chat are probably the most special/confusing, the others are pretty much requests and responses with no order, please feel free to include that somewhere or even a new page describing that flow

ericdallo: @bombaywalla about the integration-tests: It seems some order [was changed indeed](https://github.com/editor-code-assistant/eca/actions/runs/17351191942/job/49257740005?pr=104#step:6:3163), the tests are easy to understand tho, you can build locally with `bb debug-cli` and then `bb integration-test`, then you can follow the messages and understand what's different.

bombaywalla: I ran the two commands you suggested on the master branch.
And those tests failed as well. MacOS 15.6.1 Arm M4, in case that matters.
I get the following:

```
Ran 14 tests containing 295 assertions.
17 failures, 2 errors.
```


What might I be doing wrong?  I'd have expected these tests to pass.

ericdallo: Oh fail in master is not good, I will check them

ericdallo: i will take a closer look in a few hours, but the tests are passing on [master](https://github.com/editor-code-assistant/eca/actions/runs/17348495302)

bombaywalla: That's a relief.  So, it must be something in my environment.  Thanks for checking.
I'll see if I can narrow down the issue on my end.

bombaywalla: I'm still getting errors.  On MacOS 15.6.1 M4.
Would you see if the following commands run without error in your env?
I just ran `bb debug-cli` followed by `bb integration-test > test.out 2> test.err` in  a freshly cloned repo.
Is there anything else you can spot that I am doing wrong?

```sh
(base) dorab@Dorabs-Mac-mini ~ % cd Projects/testeca 
(base) dorab@Dorabs-Mac-mini testeca % rm -rf eca
(base) dorab@Dorabs-Mac-mini testeca % git clone https://github.com/editor-code-assistant/eca.git
Cloning into 'eca'...
remote: Enumerating objects: 13289, done.
remote: Counting objects: 100% (1102/1102), done.
remote: Compressing objects: 100% (215/215), done.
remote: Total 13289 (delta 834), reused 937 (delta 746), pack-reused 12187 (from 3)
Receiving objects: 100% (13289/13289), 7.85 MiB | 25.53 MiB/s, done.
Resolving deltas: 100% (7108/7108), done.
(base) dorab@Dorabs-Mac-mini testeca % cd eca
(base) dorab@Dorabs-Mac-mini eca % bb debug-cli
Building uberjar...
Generating bin...
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by io.netty.util.internal.PlatformDependent0$3 (file:/Users/dorab/.m2/repository/io/netty/netty-common/4.1.30.Final/netty-common-4.1.30.Final.jar)
WARNING: Please consider reporting this to the maintainers of class io.netty.util.internal.PlatformDependent0$3
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
WARNING: seqable? already refers to: #'clojure.core/seqable? in namespace: clojure.core.incubator, being replaced by: #'clojure.core.incubator/seqable?
WARNING: seqable? already refers to: #'clojure.core/seqable? in namespace: clostache.parser, being replaced by: #'clojure.core.incubator/seqable?
Creating unix standalone executable: eca
(base) dorab@Dorabs-Mac-mini eca % bb integration-test > test.out 2> test.err
(base) dorab@Dorabs-Mac-mini eca % tail test.out
2025-08-31T14:11:39.545119 Client 14 received notification: {:jsonrpc "2.0", :method "chat/contentReceived", :params {:chatId "91003039-7738-4144-84f0-03f022107be5", :requestId 2, :role "assistant", :content {:type "text", :text "Foo"}}}
2025-08-31T14:11:39.545406 Client 14 received notification: {:jsonrpc "2.0", :method "chat/contentReceived", :params {:chatId "91003039-7738-4144-84f0-03f022107be5", :requestId 2, :role "assistant", :content {:type "text", :text " bar!"}}}
2025-08-31T14:11:39.545746 Client 14 received notification: {:jsonrpc "2.0", :method "chat/contentReceived", :params {:chatId "91003039-7738-4144-84f0-03f022107be5", :requestId 2, :role "assistant", :content {:type "text", :text "\n\n"}}}
2025-08-31T14:11:39.545948 Client 14 received notification: {:jsonrpc "2.0", :method "chat/contentReceived", :params {:chatId "91003039-7738-4144-84f0-03f022107be5", :requestId 2, :role "assistant", :content {:type "text", :text "Ha!"}}}
2025-08-31T14:11:39.546082 Client 14 received notification: {:jsonrpc "2.0", :method "chat/contentReceived", :params {:chatId "91003039-7738-4144-84f0-03f022107be5", :requestId 2, :role "system", :content {:type "progress", :state "finished"}}}
2025-08-31T14:11:40.049865 Client 14 lifecycle: shutting down
2025-08-31T14:11:40.050258 Client 14 lifecycle: shutdown

Ran 14 tests containing 295 assertions.
17 failures, 2 errors.
(base) dorab@Dorabs-Mac-mini eca % java -version
openjdk version "24.0.2" 2025-07-15
OpenJDK Runtime Environment Temurin-24.0.2+12 (build 24.0.2+12)
OpenJDK 64-Bit Server VM Temurin-24.0.2+12 (build 24.0.2+12, mixed mode, sharing)
(base) dorab@Dorabs-Mac-mini eca % 
```

bombaywalla: I also tried it with `.` explicitly first in my `$PATH` and still get the same errors.

ericdallo: @bombaywalla I'm sorry, there was indeed a error related to a recent change of mine, I don't know why it wasn't caught before, just fixed.

bombaywalla: Thanks.  I can now run integration tests (on master) without any error.  Now to see why it fails on my branch.

src/eca/features/chat.clj
```diff
@@ -42,12 +42,238 @@
     (db/update-workspaces-cache! @db*)))
 
 (defn ^:private assert-chat-not-stopped! [{:keys [chat-id db*] :as chat-ctx}]
+  ;; TODO: Should :stoping be :stopping?
```
ericdallo: both works for me

src/eca/features/chat.clj
ericdallo, line 69: Would be nice to have good unit tests for this state machine, I believe you have that in mind

src/eca/features/chat.clj
bombaywalla: As far as I am aware, "stopping" is the correct spelling.
If you are okay with it, I can make the change "stoping" -> "stopping" throughout.

src/eca/features/chat.clj
bombaywalla: Agreed.  Working on it.

bombaywalla: Turns out that the integration tests now pass, without any further changes on my branch.
Thanks for fixing the issue.

src/eca/features/chat.clj
ericdallo: ahh, sorry, I thought you meant the keyword with the `?` hehe, yeah, my grammar is not the best hehe, thanks

src/eca/features/chat.clj
```diff
@@ -42,12 +42,237 @@
...
 
 (defn ^:private assert-chat-not-stopped! [{:keys [chat-id db*] :as chat-ctx}]
-  (when (identical? :stoping (get-in @db* [:chats chat-id :status]))
+  (when (identical? :stopping (get-in @db* [:chats chat-id :status]))
     (finish-chat-prompt! :idle chat-ctx)
     (logger/info logger-tag "Chat prompt stopped:" chat-id)
     (throw (ex-info "Chat prompt stopped" {:silent? true
                                            :chat-id chat-id}))))
 
+;;; Helper functions for tool call state management
```
ericdallo, line 50: do we need this kind of line comment?

src/eca/features/chat.clj
bombaywalla: Unit tests completed.

test/eca/features/chat_tool_call_state_test.clj
```diff
@@ -0,0 +1,944 @@
+(ns eca.features.chat-tool-call-state-test
```
ericdallo, line 1: I usually prefer test ns to follow the same ns name, and have deftests per function being tested with multiple testings, unless too complex (which looks like this one). Wondering if we should not move those tests to eca.features.chat-test

bombaywalla: There are a lot of unit tests for this state management code, so I kept them separate.
Personally, I would keep them separate.  But if you prefer them in `chat_test.clj` I can do that.
Let me know your decision.

src/eca/features/chat.clj
bombaywalla: We don't **need** it, but I find it useful to separate the helpers from the main code.
I can remove if you prefer.

test/eca/features/chat_tool_call_state_test.clj
ericdallo: Ok, let's keep it here and It becomes weird we move to chat later.

src/eca/features/chat.clj
ericdallo: yeah, I wonder if that doesn't show how this bunch of logic should not belong to this ns maybe, but seems to overcomplicate it

bombaywalla: There are a total of 2 helper functions (used elsewhere in the file as well), one state machine definition, one helper function used only for state management and one main function that does the state transition management.  I don't consider that complicated.  MY preference is to keep these functions in `chat.clj`, but I am happy to remove the comment or move the functions to another ns if that is your preference.  Just let me know.

ericdallo: Yeah, I meant the whole PR is refactoring this logic and has lots of code related to it, but I understand it's a feature that is on a complex stage (complex meaning needs a state machine).
It's ok for now, let's see how this evolves in the future

test/eca/features/chat_tool_call_state_test.clj
bombaywalla: OK, thanks.

Would you please take a look at master?
The integration tests are failing again on github and also in my repo after updating the latest github master.
The integration tests are failing in the same way as the last time.
I have only added tests, not changed the source code.
So, I'm guessing the problem is the same as last time?
Please check and let me know.

Thanks for your help.

src/eca/features/chat.clj
bombaywalla: Okay.

Would you please particularly review the state machine and confirm that the state transitions and actions on the :stop-requested event are what you desire.  Thanks.

test/eca/features/chat_tool_call_state_test.clj
ericdallo: @bombaywalla integration tests at master are now passing (tested local as well), so I believe it's related with your PR or some merge with master?

bombaywalla: Thanks for checking.

I can't quite figure out what is wrong, since I only added unit tests and did not change code.  I did `git push origin --force-with-lease` to push the branch.  That's the only thing I can currently think of that causes this problem.  Anyways, I will investigate more.

ericdallo: Maybe try merging master again? I can take a look tomorrow if you don't find it.

bombaywalla:
I updated this PR to handle the changes made to master today.
All tests pass.
This PR is ready to review.

test/eca/features/chat_tool_call_state_test.clj
bombaywalla: It worked after merging and rebasing off the latest changes to master today.

src/eca/features/chat.clj
```diff
@@ -42,12 +42,240 @@
...
+
+   [:preparing :tool-prepare]
+   {:status :preparing
+    :actions [:send-toolCallPrepare]} ; Multiple prepares allowed
+
+   [:preparing :tool-run]
+   {:status :check-approval
+    :actions [:init-approval-promise :send-toolCallRun]}
+
+   [:check-approval :manual-approve]
```
ericdallo: Can we do this minor change to match new naming in toolCall approval.
```suggestion
   [:check-approval :ask-approve]
```

src/eca/features/chat.clj
```diff
@@ -42,12 +42,240 @@
...
+   {:status :waiting-approval
+    :actions [:send-progress]}
+
+   [:check-approval :auto-approve]
```
ericdallo: Can we do this minor change to match new naming in toolCall approval.
```suggestion
   [:check-approval :allow-approve]
```

src/eca/features/chat.clj
```diff
@@ -235,17 +459,15 @@
...
+                                                  (transition-tool-call! db* chat-ctx id :execution-end
+                                                                         {:origin origin
+                                                                          :name name
+                                                                          :arguments arguments
+                                                                          :error (:error result)
+                                                                          :outputs (:contents result)
+                                                                          :details details
+                                                                          :summary summary})))
+                                              ;; assert: In :rejected state
                                               (let [[reject-reason reject-text] (if (= :deny approval)
```
ericdallo: now that we support `:deny` configs in tool call, they can be rejected by user choice in chat like before + user deny config, should we support a reject state for that or is it ok to keep current ones?

src/eca/features/chat.clj
bombaywalla: Sure.  Makes sense.  I will make the change today.

src/eca/features/chat.clj
bombaywalla: Sure. Makes sense.  Let me look at it and make the change later today.

src/eca/features/chat.clj
bombaywalla: There is already a :rejected state.  Let me take a look later today and see if an additional transition can be easily added.

bombaywalla:
Manual testing is likely a wise move in this situation.
I will review later today.

ericdallo:
@bombaywalla just got a invalid transition testing:


It seems it went from a approved -> rejected somehow ? BTW this tool was in my `toolCall approval deny` config, so it may be related with recent master changes you merged yesterday.

bombaywalla:
Thanks for reporting.  Yes, most likely caused by a poor integration of yesterday's changes to master into my branch.  I will take a look later today.

bombaywalla:
Hi.  It looks like you merged this branch into master already.  So, I am not sure of what convention/protocol to follow for continued development?  Should I just create a new branch (and thus new PR) and work off the latest master?  Or do I continue to make changes to this branch and PR?  I think the former would be cleaner.  How would you like me to proceed?

bombaywalla:
Oops I think I misread the merge message.  It seems you merged master into this branch - the opposite of what I thought.  So no worries, I will continue on this branch.

src/eca/features/chat.clj
bombaywalla: How about `:default-approve` (my preference) or `:config-approve` instead of `:auto-approve`?
I think it more accurately describes what is going on.

ericdallo: I think I thought on `allow-approval` actually, which matches the names on configs

bombaywalla: Okay.  I will use that.

bombaywalla: Sorry, on second thoughts, let me explain what is going on.
And why I think `:default-approve` is still a better choice.

Currently, when you are in `:check-approval` state and you get an `:auto-approve` event, it will then transition to `:execution-approved` (which is incorrect and the cause of the invalid state transition you found), and then deliver whatever the default config setting is other than `:ask`.  Meaning that if the config setting is `:allow` it will deliver `true` to the promise and if the config setting is `:deny`, then it will deliver `false`.  So, if we use an event name of `:allow-approval`, that does not not accurately reflect what is happening.

That being said, I'm still thinking about the best way to handle the invalid state transition you found.
The actions are correct, but the next-state is wrong.

I hope I have explained the situation properly.
Let me know your thoughts.

bombaywalla: In order to fix the invalid state transition, I came up with a better (and more correct) way of handling these choices.  I should have that ready soon - just updating the tests.  Then you can review and let me know if you want changes.

bombaywalla:
The latest commit addresses:
1. The invalid state transition bug you found.  Changed the state machine slightly.
2. The name changes you requested.
3. Handling the :rejected state that can be reached via a :config-deny or :user-reject.
4. Tests have been updated.
Please review and test manually.
Let me know if further changes are desired.

src/eca/features/chat.clj
ericdallo: That makes sense!

bombaywalla:
A very minor point.....
the `:on-tools-called` callback key should be `:on-tool-called` for consistency with the other keys.
If you agree, I can make that change too.

ericdallo:
> the :on-tools-called callback key should be :on-tool-callback for consistency with the other keys.

@bombaywalla not sure I got this, what other keys have this standard? actually they represent what LLM did, and in this case, it called a tool/s

Thanks for the changes, I'll probably test/erge first #105 and then test yours, may be done today night/tomorrow

bombaywalla:
All the other callback keys use "tool" as singular, except for this one, which uses "tools" as plural.
Also, the name of the protocol message related to this callback is "toolCalled" not "toolsCalled" so that's another reason to use the singular.

I had a typo (since corrected) in my previous comment which likely confused things.  Sorry.

ericdallo:
@bombaywalla the point is that LLM can call multiple tools in a single response, that's why we have a loop to spawn multiple toolCallPrepare/run/called to client UI, so the callback in plural is intended

bombaywalla:
Okay. Thanks.  I removed the comment from the file and updated the PR.

ericdallo:
@bombaywalla overall looks good, but I tested something from #65 and I'm not sure if you added the states to fix it:



I just stopped the prompt while waiting for a tool call approval

bombaywalla:
Thanks for that.  I see the issue.  There is currently a transition from `:waiting-approval` on the event `:stop-requested`.  The problem is, what should be the proper terminal state.  I will take a closer look later today and get back.

bombaywalla:
I have a work-in-progress checkin.

There are still some issues to handle.
In particular, there are concerns that the approval promise will not
be derefed if we get a :stop-requested and will cause a memory leak.
I have not updated the tests, so they will fail.

Also, tests have not been updated yet, and so will fail.  I feel that the current tests
are too brittle and they need to be changed or removed.  Maybe you had suspected that all along :-)

You had recently added some code to record whether the reject was because of a user action or because of a config.
The only place this is used outside the `:on-tools-called` callback is that the information is sent as the `:reason` in a `toolCallReject` notification.

As I was adding the code to reject the call because of the stopping of the prompt, I needed to add another reason for denying the tool call.
I have added some code to record the reasons for the decision to either allow or deny a tool call.
You only had deny reasons and thought it might be good to also record (if not use) the allow reasons.
In that case, it might be good to make a small breaking change to the `:reason` key in the `toolCallReject` notification.
From `:user-choice` to `:user-choice-deny`.  That would allow us to introduce codes that also send the reasons for allow.
I also checked the protocol spec (which just says it should be the string 'user') and it does not seem to match the code.
Since this was a recent change, I believe, I thought it might be okay to make this slightly breaking change.
Are you okay with that?  If not, no problem - I can adapt to whatever you want.
I have not changed any of the code regarding this at the moment, so the current functionality is not impacted.

bombaywalla:
I also mean to ask you whether you prefer that I `git rebase` or `git merge` my branch from your master when I want to update my branch with the changes made to master?  So far I have been using `rebase`, but either is fine with me.

ericdallo:
@bombaywalla sorry for the recent changes making it harder to merge this PR.

Yes, the breaking change is ok, since I know all clients and they are not using that reason yet.

I think merge with master is better, since I can review specific commits only done after my last review

bombaywalla:
Latest revision.
master is merged.
Tests pass locally.

I want to also run manual tests.
How do I do that?
`bb debug-cli` and then run that `eca` binary?

bombaywalla:
I was able to test the :waiting-approval -> :rejected state transition manually using eca-emacs, and it works.

Please review the state machine transitions for the :stop-requested event and verify that those are the transitions you intended.

This PR is ready for review.

ericdallo:
@bombaywalla I will test right now!

I wanna add a new toolCallRunning state between toolCallRun and toolCalled, so I wanna merge this PR first to avoid conflicts.

ericdallo:
Tip: if using linux, you can download the compiled graalvm build from the [CI summary](https://github.com/editor-code-assistant/eca/actions/runs/17473719575)

ericdallo (approved):
@bombaywalla Looks good and now everything seems to be working, I will try to add the new `toolCallRunning` (since we don't have a state of when the tool call is running effectively, only when it's gonna start to run but may take seconds/mins) and let you know how I think the state machined helped with that.

ericdallo:
@bombaywalla just noticed we are returning rejected too calls with null fields when stopping a prompt:
```javascript
Params: {
  "chat-id" : "7ec9b2bf-868d-482e-9288-f39d15ebed47",
  "role" : "assistant",
  "content" : {
    "type" : "toolCallRejected",
    "id" : "call_09c8td6tKjYvtw6bho1fkSUK",
    "origin" : null,
    "name" : null,
    "arguments" : null,
    "reason" : "user-prompt-stop"
  }
}
```

do you think we can fill those? otherwise I will have to change clients to handle those but I'd like to avoid it.

bombaywalla:
I will take a look and report back.

ericdallo:
FYI @bombaywalla just added `toolCallRunning` new state and it was easy peasy with the state machine, thanks!

bombaywalla:
So, I looked at this, and from my limited understanding, the crux is that `prompt-stop` does not, and can not, have the missing parameters needed in the `toolCallReject` call.  If the `toolCallReject` needs these parameters, I think we will have to store them in the `db*` atom in the tool call state.  As far as I am aware, the tool call state currently only has `:status` and `:approval*` in it.  We could easily add the parameters there too.  Please correct me if I have misunderstood.  If you think this  is a good approach, I am happy to make the changes in a new PR if you want.  Let me know.

ericdallo:
Yep, that's what I thought as well, I think it's ok to add those to the toolCall atom in the chat.clj as long we clean that after the toolCall reached a finished state.

bombaywalla:
Okay, I'll look into implementing this.
Out of curiosity, why clean up just the parameters after the tool Call?
I don't believe we are currently cleaning up the `:status` and `:approval*` at the current time.

bombaywalla:
I took a look at implementing this.
In the investigation, I came across a potential problem/inconsistency.

In the protocol spec for `toolCallPrepare` I see that there is a parameter `argumentsText: string;`.
In the spec for `toolCallRejected` I see that there is a parameter `arguments: {[key: string]: string};`.
Why does `toolCallPrepare` take a different arguments parameter than `toolCallRejected`?
Why not have `toolCallPrepare` also take `arguments: {[key: string]: string};`?
I was hoping to use the arguments provided to `toolCallPrepare` to save the parameters that could then be used in sending `toolCallRejected`, but it seems that the arguments are different.

bombaywalla:
I looked at the code some more, specifically `anthropic.clj`, and I see that what we are hoping for may not be possible in general.
It seems that the `argumentsText` in the `toolCallPrepare` is the possibly incomplete JSON encoded string fragment that represents the arguments.  The only time we have the complete arguments is in the `toolCalled` protocol message.
So, if we get a `prompt-stop` after a `toolCallPrepare` but before a `toolCalled`, there is no way we have the arguments.  So, in general, the clients will have to be changed to allow the arguments in `toolCallReject` to be null.
However, I can fill in the `:name` and `:origin` on `toolCallPrepare`, and fill in the `arguments` on `toolCalled`.
Is my understanding correct?

If so, are you okay proceeding as described above?

ericdallo:
Sorry for the late but I think you figured it out by yourself, and yes, you are right, I was aware of that and I'm ok fixing for name and origin but not arguments if not possible

bombaywalla:
No worries.  See #109 for a PR on this.

PR #105, opened 2025-09-02
custom tools support

CsBigDataHub:
@ericdallo, please review this draft. I will update documentation once we are good with this implementation. 

#92


src/eca/config.clj
```diff
@@ -79,6 +79,12 @@
            :repoMap {:maxTotalEntries 800
                      :maxEntriesPerDir 50}}})
 
+(defn custom-tools
```
ericdallo: We should not create new methods here, we have all config already in `all` function.
Especially getting from initial config which is wrong.

src/eca/features/tools.clj
```diff
@@ -19,6 +21,43 @@
...
+    {(name tool-name)
+     {:name (name tool-name)
+      :description (:description tool-config)
+      :parameters {:type "object"
+                   :properties (update-keys (:properties schema) keyword)
+                   :required (mapv keyword (:required schema))}
+      :handler (build-tool-fn tool-config)}}))
+
+(defn- load-custom-tools
+  "Loads all custom tools from the config, transforms them, and merges them into a single map."
```
ericdallo: This kind of docstring does not helps with anything IMO.

src/eca/features/tools.clj
```diff
@@ -19,6 +21,43 @@
...
+  (let [schema (:schema tool-config)]
+    {(name tool-name)
+     {:name (name tool-name)
+      :description (:description tool-config)
+      :parameters {:type "object"
+                   :properties (update-keys (:properties schema) keyword)
+                   :required (mapv keyword (:required schema))}
+      :handler (build-tool-fn tool-config)}}))
+
+(defn- load-custom-tools
```
ericdallo: Can we move this and functions above to a f.tools.custom ns?
Following the same of other tools features

ChetanKoneru:
@ericdallo , How is this? I hope this is much cleaner?

src/eca/features/tools/custom.clj
```diff
@@ -0,0 +1,43 @@
+(ns eca.features.tools.custom
+  (:require
+   [babashka.process :as process]
+   [clojure.string :as string]))
+
+(set! *warn-on-reflection* true)
+
+(defn- build-tool-fn
```
ericdallo:
```suggestion
(defn ^:private build-tool-fn
```

src/eca/features/tools/custom.clj
```diff
@@ -0,0 +1,43 @@
...
+                                (let [key-name (keyword (subs part 2 (- (count part) 2)))]
+                                  (str (get llm-args key-name "")))
+                                part))
+                            command)
+          {:keys [out exit]} (process/sh resolved-command {:error-to-out true})]
+      (if (zero? exit)
+        out
+        (str "Error: Command failed with exit code " exit "\nOutput:\n" out)))))
+
+(defn- custom-tool->tool-def
```
ericdallo:
```suggestion
(defn ^:private custom-tool->tool-def
```

ericdallo:
While I test, can you update CHANGELOG too @CsBigDataHub ?

ChetanKoneru:
Sure, will it be 0.44.2 or 0.45.0

ericdallo:
you add under unreleased @CsBigDataHub

ericdallo:
@CsBigDataHub There are still some issues, like wrong config (`custom-tools` instead of `customTools`), wrong output from tool call, but those should be easy to fix, I will merge and fix myself, thanks for your help!

PR #108, opened 2025-09-04
feat: add nix flake build

### LINKED ISSUE #106: nix/flake build

It would be nice to be able to install this application on nix. It can be either a packaged nix application, or a flake. 

Actually I tried to build a flake for this repo by myself, but it failed because the buildPhase has no network access to fetch clojure's git dependencies.

sunng87:

Fixes #106 

This patch follows what we discussed in #106 to include a nix flake build file and CI task. The nix flake is mostly copied from clojure-lsp. Because the flake build requires a deps-lock.json as lockfile of clojure dependencies, I include a ci task to ensure it's always up to date.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,64 +4,77 @@
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
     flake-utils.url = "github:numtide/flake-utils";
+    clj-nix = {
+      url = "github:jlesquembre/clj-nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
   };
 
-  outputs = { self, nixpkgs, flake-utils }:
+  outputs = { self, nixpkgs, flake-utils, clj-nix }:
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -0,0 +1,80 @@
...
+          in
+            pkgs.mkShell
+              {
+                packages = [ deps-lock-update ];
+                nativeBuildInputs = with pkgs; [
+                  babashka
+                  graalvm
+                  jdk
+                  clojure
+                  clojure-lsp
```
ericdallo, line 41: I think we don't need this?

sunng87: I'm using envrc with nix flake so when I'm working on the project these devShell native dependencies are activated by default. So I often include essential dev tools here, like this https://github.com/sunng87/pgwire/blob/master/flake.nix#L31

But I'm open to your opinions.

ericdallo: I'm ok with that I guess

ericdallo:
Thank you!

PR #109, opened 2025-09-06
Add missing parameters to toolCallRejected

bombaywalla:
As we discussed after PR #104 was merged, `toolCallRejected` did not populate some of the values on `prompt-stop`.

It seems that the `argumentsText` in the `toolCallPrepare` message is the possibly incomplete JSON encoded string fragment that represents the arguments.
The only time we have the complete arguments is in the `:on-tools-called` callback.
So, if we get a `prompt-stop` after a `toolCallPrepare` but before a `toolCalled`, there is no way we have the arguments.
Therefore, in general, the clients will have to be changed to allow the `:arguments` in `toolCallRejected` to be null.
However, we can fill in the `:name` and `:origin` on `toolCallPrepare`, and fill in the `:arguments` on `toolCalled`.
- [ x] I added a entry in changelog under unreleased section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Added missing parameters to `toolCallRejected` where possible.  PR #109
+
 ## 0.49.0
 
 - Add `totalTimeMs` to reason and toolCall content blocks.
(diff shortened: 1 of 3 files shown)


ericdallo (approved):
Thank you!

PR #118, opened 2025-09-13
Improve large file handling in `read-file` tool

bobrowadam:
- Replace basic truncation notice with detailed line range information and next-step instructions.
- Allow users to customize default line limit through `tools.readFile.maxLines` configuration (keep the current 2000 as default).

### DIFF
--- a/src/eca/config.clj
+++ b/src/eca/config.clj
@@ -84,7 +84,8 @@
                                  "eca_grep" {}
                                  "eca_editor_diagnostics" {}}
                          :ask {}
-                         :deny {}}}
+                         :deny {}}
+              :readFile {:maxLines 2000}}
    :mcpTimeoutSeconds 60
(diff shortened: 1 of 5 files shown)


src/eca/config.clj
```diff
@@ -76,6 +76,7 @@
    :rules []
    :commands []
    :disabledTools []
+   :tools {:readFile {:maxLines 2000}}
```
ericdallo: We don't have this `tools` section anymore, can we move it to `toolCall`?

ericdallo (approved):
Looks good, can you also change the default and spec config at configuration.md?

src/eca/config.clj
bobrowadam: Sure. I made a change. Let me know if that's what you meant.

ericdallo (approved):
Awesome, thank you!

Issue #111, opened 2025-09-07
promptStop should cancel already executing tool call futures


bombaywalla:
To upvote this issue, give it a thumbs up. See [this list](https://github.com/editor-code-assistant/eca/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

When a `chat/promptStop` is executed while a tool call is executing in its own future, we currently get an "Invalid state transition" (by design).  ECA then waits for the future to complete.  It would be good if the future could be cancelled - especially if the tool call is long-lived.  However, it may be challenging since the future thread has to cooperate with the cancellation process.  And we don't always have control over the implementation.

Steps to reproduce the behavior (in eca-emacs):
1. Send a prompt like `Please run the bash tool with a sleep of 10 secs followed by a 'ls'.`.
2. Click on `accept` to allow the tool.
3. Click on `stop`, to stop the prompt, while the tool is still running.
4. Check the `<eca:stderr:1>` buffer.  You should see an "Invalid state transition" exception.

What we'd like in this issue is just to be able to cancel the future - ideally in a clean way, but okay if not clean.
Secondarily, we want to allow the user to interrupt ECA from hanging on a long-lived call, but see #77 on background tasks and long-lived processes.
Minimally, get rid of the invalid state transition exception.

bombaywalla:
Are you okay with storing the futures in the tool call state?

ericdallo:
@bombaywalla I guess so, we should just:
- make sure we clean it to avoid memory leak after the tool was executed
- make sure we don't serialize it to a file, but I believe we already do this, check [here](https://github.com/editor-code-assistant/eca/blob/ba958ea368abe1360be5d70411e0399d57bb670c/src/eca/db.clj#L126)

bombaywalla:
I checked the place you pointed me at and yes, we are `dissoc`ing all the tool calls before serializing to a file.

Do you want me to remove the whole tool call state after the tool finishes executing or is stopped?  Or just the future?
Should the cleanup be done as soon at the tool call completes, or after the chat ends?

What are your thoughts about supporting "background" tool calls, as suggested in #77 ?
If we do that, we may need to keep some state around to note that the tool is running?
I'd prefer to come up with a solution to this issue that will be compatible with a "background" tool call if/when it is implemented.

Right now, I'm in the "hammock" phase :-) so just thinking about potential pros/cons of different approaches.

ericdallo:
[...]

PR #119, opened 2025-09-14
Moved the future in :on-tools-called and stored it in the db.

bombaywalla:
The future in the `:on-tools-called` callback used to be just before the check for `@approved*`.  There is really no need to have a future when the call has been rejected, only when the call has been allowed and ready to go.  So, the future was moved to just before the call.

To prepare for being able to cancel the call future, the future needs to be stored in the db.  This is done by a new action `set-call-future` that is called on the `:execution-start` event. However, the since the future starts running as soon as it s created, we have to first delay the future and then force it only after the tool call state has been transitioned to `:executing`.

There should be no change in external behavior.

### DIFF
--- a/src/eca/features/chat.clj
+++ b/src/eca/features/chat.clj
@@ -56,7 +56,7 @@
   (get-in db [:chats chat-id :tool-calls tool-call-id]))
 
 (defn ^:private get-active-tool-calls
-  "Returns a map of tool calls that are still active.
+  "Returns a map of tool-call-id -> tool calls that are still active.
 
   Active tool calls are those not in the following (terminal) states: :completed, :rejected, :stopped."
   [db chat-id]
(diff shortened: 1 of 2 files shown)


bombaywalla:
I would like to run a manual test before making this PR ready for review.

bombaywalla:
This PR is a part of Issue #111 and is a prerequisite for what is to follow

bombaywalla:
I was just hoping to do a manual test, but it's ready to review otherwise.

bombaywalla:
I've done some manual testing.
Please check it manually yourself as well.
This PR could do with more rigorous manual testing.

ericdallo:
Thank you, will review/test now

ericdallo (approved):
Works!

PR #132, opened 2025-09-26
Add :stopping status. Pass state fns to tool calls.

bombaywalla:
This is an interim commit to eventually allow tool calls to be interrupted.

In preparation, we added a :stopping status in between :executing and :stopped.  This allows us to have self transitions to create and destroy tool call resources.  The transitions and events are there, but currently they are not being exercised.  The :stopped status is considered an "inactive" status similar to an initial or terminal status.

There is one semantic behavioral change.  Previously, when a tool call was :executing and it got a :stop-requested event, that was an invalid transition.  Now, the transition happens and after the tool call is completed, a toolCallReject message is sent to the client.  Note that ECA will still wait for the tool call to complete and will not interrupt it - yet.  That is for a subsequent change.

To allow a tool call to update the tool call state, we have added two functions that are passed along to the tools.  One function retrieves the current state of the tool call and the other updates it. Currently, these functions are passed along to lower level functions but are not used.

Added tests for the new status and transitions.

### DIFF
--- a/src/eca/features/chat.clj
+++ b/src/eca/features/chat.clj
@@ -58,11 +58,11 @@
 (defn ^:private get-active-tool-calls
   "Returns a map of tool-call-id -> tool calls that are still active.
 
-  Active tool calls are those not in the following (terminal) states: :completed, :rejected, :stopped."
+  Active tool calls are those not in the following states: :completed, :rejected, :stopping, :stopped."
   [db chat-id]
   (->> (get-in db [:chats chat-id :tool-calls] {})
        (remove (fn [[_ state]]
-                 (#{:completed :rejected :stopped} (:status state))))
+                 (#{:completed :rejected :stopping :stopped} (:status state))))
        (into {})))
 
 ;;; Event-driven state machine for tool calls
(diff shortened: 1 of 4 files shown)


bombaywalla:
Unit tests pass.
I would like to run a manual test, but other than that, this PR should be ready to review.
Please also run a manual test yourself to see that there are no regressions.
Thanks

src/eca/features/chat.clj
```diff
@@ -141,10 +159,37 @@
...
 
+   [:executing :resources-created]
+   {:status :executing
+    :actions [:add-resources]}
+
+   [:executing :resources-destroyed]
+   {:status :executing
+    :actions [:remove-resources]}
+
+   ;; I don't think this transition is needed
```
ericdallo, line 170: Agree, I think we can remove it.

src/eca/features/chat.clj
```diff
@@ -141,10 +159,37 @@
    {:status :completed
     :actions [:send-toolCalled :log-metrics :send-progress]}
 
+   [:executing :resources-created]
```
ericdallo, line 162: What you mean with resource here? You mean the future or the process of the tool? that's what you call resource?

ericdallo (approved):
Besides my question it looks good

src/eca/features/chat.clj
bombaywalla: Right.  By "resource" I mean any stateful part of the tool call itself that needs to be reclaimed on cancellation (or completion).  For example, in the shell tool, a resource might be the handle on the process that is spawned.  In an MCP tool, it might be the MCP client.  In general, a tool may need multiple resources to accomplish its tasks.

I haven't quite decided on whether the future should be a resource or not.  Mainly because the future handle is the responsibility of the calling thread and not of the tool call itself.  In any case, even if it isn't a resource, it behaves similarly - only in the parent thread.  My current feeling is that the future is not a tool call resource.

bombaywalla:
I have also run a  quick manual test and verified that when I stop a running call it does not cause an invalid transition exception now, and eventually rejects the call.

Ready to merge if you're okay with it.

src/eca/features/chat.clj
ericdallo: Got it, sounds to me something more like a "state", but I'm ok with "resource" as well.

PR #133, opened 2025-09-27
Allow parsing zero or more whitepsace after data:

jilen:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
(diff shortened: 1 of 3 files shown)


jilen:
Some provider dont' have whitespace after `data:`.

ericdallo (commented):
Hum need to test that carefully, and check if there is any corner case, a unit test for that would be nice too

jilen:
Updated to trim leading whitespace only, seems much safer.

ericdallo:
Thank you!

PR #136, opened 2025-09-29
Fix NullPointerException in native-definitions by Guarding :description Replacement

CsBigDataHub:
This PR addresses a server-side NullPointerException thrown from clojure.string/replace when converting tool definitions in eca.features.tools/native-definitions. The exception occurred because some tool entries had a nil :description, and string/replace cannot accept nil as input. We now ensure only actual strings are processed, leaving nil or non-string values untouched.
Background

Issue: When users changed chat behavior or sent prompts, the backend crashed with:

see - https://github.com/editor-code-assistant/eca-emacs/issues/46

```
java.lang.NullPointerException
  at clojure.string$replace.invokeStatic(string.clj:101)
  ...

Cause: In native-definitions, every tool map’s :description was unconditionally passed to string/replace to substitute $workspaceRoots. If :description was nil, this thrown an NPE.
```

Impact: All Emacs clients experienced chat failures on any tool operation requiring native definitions.

### DIFF
--- a/src/eca/features/tools.clj
+++ b/src/eca/features/tools.clj
@@ -44,10 +44,13 @@
    (map (fn [[name tool]]
           [name (-> tool
                     (assoc :name name)
-                    (update :description #(if (string? %)
-                                           (string/replace % #"\$workspaceRoots" 
-                                                          (constantly (tools.util/workspace-roots-strs db)))
-                                           %)))])
+                    (update :description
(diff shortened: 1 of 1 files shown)


ericdallo:
Actually, it's missing a paren

ericdallo:
Closing as mentioned in the issue

PR #138, opened 2025-09-30
Add github-copilot/claude-sonnet-4.5

deejayem:
Add Copilot has Sonnet 4.5 now, so I've added it to the available models.

Does this need a changelog entry?

Hopefully I didn't break anything: the unit tests work for me, and I got four integration test failures before my change, and only two after.


ericdallo:
Thanks for the PR, but I did that this afternoon and checked the PR only now, so this is already implemented, thank you anyway!

PR #146, opened 2025-10-06
Add support for secrets stored in authinfo and netrc files

r0man:
I would like to mamange my ECA config and commands with Guix and don't want any secrets in those files. I could use keyEnv for this, but all my other secrets are already in `~/.authinfo.gpg` which I would like to use.

This PR adds support for reading secrets from the `~/.authinfo` and `~/.netrc` files. It supports GPG encrypted and Windows variants.


plans/authinfo-netrc-secrets.md
```diff
@@ -0,0 +1,1523 @@
+# Plan: Authinfo and Netrc Secrets
```
ericdallo: why we need to commit this file?

resources/META-INF/native-image/eca/eca/native-image.properties
```diff
@@ -7,7 +7,7 @@ Args=-J-Dborkdude.dynaload.aot=true \
      --initialize-at-build-time=org.yaml.snakeyaml \
      --initialize-at-build-time=io.opentelemetry.api.common.AttributeType \
      --initialize-at-build-time=io.opentelemetry.api.logs.Severity \
-     --initialize-at-build-time=steffan_westcott.clj_otel.context$reify__3173 \
+     --initialize-at-build-time=steffan_westcott.clj_otel.context \
```
ericdallo: This will likely break otel integration, this is something that needs to be fixed on clj-otel, and meanwhile we need to specify the reify object that can change anytime clj-otel bumps its version, a PITA indeed.

src/eca/secrets.clj
```diff
@@ -0,0 +1,280 @@
...
+  "Returns ordered list of credential file paths to check."
+  []
+  (let [home (System/getProperty "user.home")
+        files [".authinfo" "_authinfo" ".netrc" "_netrc"]]
+    (mapcat (fn [filename]
+              (let [path (str (io/file home filename))]
+                [(str path ".gpg") path]))
+            files)))
+
+(defn- file-exists?
```
ericdallo: Please match project codestyle (^:private instead of defn-)

docs/models.md
```diff
@@ -50,6 +50,171 @@ Example:
...
+    }
+  }
+}
+```
+
+**keyRc Format:**
+
+The `keyRc` value supports multiple formats for flexible credential matching:
+
+- `"machine"` - Simple machine name (e.g., `"api.openai.com"`)
```
ericdallo: `api.openai.com` is a good example of machine name? It doesnt looks to me

r0man: It depends on what you have in your RC files. My authinfo.gpg contains this for example:
```
machine api.openai.com login apikey password SECRET
machine vertex.google.com login apikey password SECRET
machine api.anthropic.com login apikey password SECRET
```

ericdallo: ah got it, makes sense

resources/META-INF/native-image/eca/eca/native-image.properties
ericdallo: to fix this one we need to replace with `reify__3312`

r0man: ok, I see. I reverted this commit again.

plans/authinfo-netrc-secrets.md
r0man: I got this from Steve Yegge's "vibe coding" podcast and looking at his Efrit Emacs package: 
- https://github.com/steveyegge/efrit/tree/main/plans

It basically serves as memory for the LLM, in case you want to further work/improve this feature. Think of it as an ADR. If you don't want this, I can remove it though.

src/eca/secrets.clj
r0man: I fixed it in the files I added. There are some other places that still use `defn-` though.

plans/authinfo-netrc-secrets.md
ericdallo: Yeah, I'm not sure I like this approach right now, I'd prefer to not commit it

r0man: ok, I removed this file.

docs/models.md
```diff
@@ -299,3 +300,30 @@ Only set this when your provider uses a different path or expects query paramete
      }
     }
     ```
+
+### Credential File Authentication
```
ericdallo: I want to have this before providers examples, to leave it as last option.

ericdallo (approved):
Codewise looks good, did you test the native binary? (You can get a static linux one in this PR GH action summary like [here](https://github.com/editor-code-assistant/eca/actions/runs/18325848036?pr=146))

resources/META-INF/native-image/eca/eca/native-image.properties
```diff
@@ -8,6 +8,7 @@ Args=-J-Dborkdude.dynaload.aot=true \
      --initialize-at-build-time=io.opentelemetry.api.common.AttributeType \
      --initialize-at-build-time=io.opentelemetry.api.logs.Severity \
      --initialize-at-build-time=steffan_westcott.clj_otel.context$reify__3173 \
+     --initialize-at-build-time=steffan_westcott.clj_otel.context$reify__3312 \
```
ericdallo:
```suggestion
     --initialize-at-build-time=steffan_westcott.clj_otel.context$reify__3325 \
```

r0man:
> Codewise looks good, did you test the native binary? (You can get a static linux one in this PR GH action summary like [here](https://github.com/editor-code-assistant/eca/actions/runs/18325976552?pr=146))

I just tried it:
```
[server] Starting server...
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by io.opentelemetry.internal.shaded.jctools.util.UnsafeAccess (file:/home/roman/Downloads/eca)
WARNING: Please consider reporting this to the maintainers of class io.opentelemetry.internal.shaded.jctools.util.UnsafeAccess
WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
```
Let me try your suggestion tweaking this reify ...

docs/models.md
r0man: Ok, I moved it

ericdallo:
@r0man all good to merge?

r0man:
No, let's wait a second. Still need to test the binary. I found something strange ...

r0man:
I think I saw some OpenAI hiccup. I tested it again and it worked fine now.

sg-qwt:
@r0man One little thing I'd suggest is to add a config to config auth-sources like emacs instead of just looking for hard coded path in home

```
(use-package auth-source
  :custom
  (auth-sources '("/run/secrets/authinfo")))
```

Also authinfo and netrc are basically the same in our use case, I'd highly suggest we just stick to one format (authinfo for example) to keep things simple and not confuse users.

r0man:
Hi @sg-qwt, 

thanks for your suggestion. I asked the AI `What is the difference between ~/.authinfo and ~/.netrc files?`:

```
Short answer:
- ~/.netrc is the long‑standing, standard credentials file used by
  many command‑line tools (ftp, curl, wget, lftp, etc.).
- ~/.authinfo is an Emacs/auth-source file; Emacs looks here (and in
  ~/.authinfo.gpg) for credentials used by Emacs packages (Gnus,
  Tramp, eww, SMTP/IMAP, etc.).

More detail:
- Who reads them:
  - .netrc: ftp, curl (--netrc), wget, and other non‑Emacs tools. They
    do not look at .authinfo.
  - .authinfo: Emacs’ auth-source. Emacs also understands .netrc.
- Format:
  - Both are “netrc-like” key–value pairs: machine HOST login USER
    password PASS …
  - .authinfo allows extra keys commonly used by Emacs (e.g., port,
    protocol, realm), and synonyms (login/user). Traditional .netrc
    recognizes mainly machine, default, login, password, account,
    macdef.
- Encryption:
  - Emacs commonly uses ~/.authinfo.gpg (GnuPG-encrypted) and decrypts
    it transparently.
  - Most non‑Emacs tools do not support encrypted .netrc files; they
    expect plain text.
- Permissions:
  - Many tools require restrictive perms (e.g., chmod 600) on
    .netrc. Good practice for .authinfo too.
- Emacs search order:
  - By default Emacs checks ~/.authinfo.gpg, then ~/.authinfo, then
    ~/.netrc.

Practical advice:
- If you need credentials for non‑Emacs tools, put them in ~/.netrc
  (using only standard netrc fields).
- If credentials are for Emacs only—or you want them encrypted—use
  ~/.authinfo.gpg.
- You can keep shared entries in ~/.netrc (Emacs will read it) and
  Emacs‑only or extra‑metadata entries in ~/.authinfo(.gpg).
```

Machine, login and password seem to be supported by both. So, I think we are fine.

Personally I use only `~/.authinfo.gpg` and that's what motivated me to contribute this feature. If you want things to be more customizable, I encourage you to open a PR for it. For me things are fine as they are.

If things are not working as expected, or if there is a bug with the current implementation, I'm happy to fix those, though.

sg-qwt:
> Hi @sg-qwt,
> 
> thanks for your suggestion. I asked the AI `What is the difference between ~/.authinfo and ~/.netrc files?`:
> 
> ```
> Short answer:
> - ~/.netrc is the long‑standing, standard credentials file used by
>   many command‑line tools (ftp, curl, wget, lftp, etc.).
> - ~/.authinfo is an Emacs/auth-source file; Emacs looks here (and in
>   ~/.authinfo.gpg) for credentials used by Emacs packages (Gnus,
>   Tramp, eww, SMTP/IMAP, etc.).
> 
> More detail:
> - Who reads them:
>   - .netrc: ftp, curl (--netrc), wget, and other non‑Emacs tools. They
>     do not look at .authinfo.
>   - .authinfo: Emacs’ auth-source. Emacs also understands .netrc.
> - Format:
>   - Both are “netrc-like” key–value pairs: machine HOST login USER
>     password PASS …
>   - .authinfo allows extra keys commonly used by Emacs (e.g., port,
>     protocol, realm), and synonyms (login/user). Traditional .netrc
>     recognizes mainly machine, default, login, password, account,
>     macdef.
> - Encryption:
>   - Emacs commonly uses ~/.authinfo.gpg (GnuPG-encrypted) and decrypts
>     it transparently.
>   - Most non‑Emacs tools do not support encrypted .netrc files; they
>     expect plain text.
> - Permissions:
>   - Many tools require restrictive perms (e.g., chmod 600) on
>     .netrc. Good practice for .authinfo too.
> - Emacs search order:
>   - By default Emacs checks ~/.authinfo.gpg, then ~/.authinfo, then
>     ~/.netrc.
> 
> Practical advice:
> - If you need credentials for non‑Emacs tools, put them in ~/.netrc
>   (using only standard netrc fields).
> - If credentials are for Emacs only—or you want them encrypted—use
>   ~/.authinfo.gpg.
> - You can keep shared entries in ~/.netrc (Emacs will read it) and
>   Emacs‑only or extra‑metadata entries in ~/.authinfo(.gpg).
> ```
> 
> Machine, login and password seem to be supported by both. So, I think we are fine.
> 
> Personally I use only `~/.authinfo.gpg` and that's what motivated me to contribute this feature. If you want things to be more customizable, I encourage you to open a PR for it. For me things are fine as they are.
> 
> If things are not working as expected, or if there is a bug with the current implementation, I'm happy to fix those, though.

Your AI answer/advice make sense, in our use case, eca is a general tool(editor agnostic) which read secrets here. As far as I'm aware, nobody outside emacs community using authinfo. And most use-case with netrc is plaintext. The emacs client is just using eca, and eca is the one needs to handle secrets. So shall we just stick to netrc in eca for standard and common practice here? 

@r0man What's your use-case of eca support authinfo here? I guess you just want shared secrets in the same authinfo file read by both eca and some emacs tools like gpt.el? But if eca is a general tool, I think we should stick to the wildly used long‑standing, standard one for compatibility?

What's your thoughts @ericdallo @r0man ?

sg-qwt:
> Preferred files are GPG-encrypted (`~/.authinfo.gpg` / `~/.netrc.gpg`); plaintext variants are supported.

Also I'm very dubious about the gpg approach as mentioned in the docs. Yes, emacs has good support for gpg. And emacs use authinfo.gpg. But here, it's not the scope of eca to handle gpg-encrypt/decrypt. Also, nowadays, many people uses age instead of gpg. In NixOS land, for example, we can have encrypted files using sops or whatever to encrypt to a netrc.age or netrc.gpg for example, then at bootup or maybe runtime decrypt that file to target plaintext `~/.netrc` for example.

Any way, my point is. The saying of `Preferred files are GPG-encrypted` are extremely weird and uncommon. And we shall just support plain text ~/.netrc just like every other tool such as curl or whatever. Drop gpg/encryption, it's not eca's duty. Encrypt secrets to a secrets store/passward manager/pain-text gpg or age file, whatever, for eca, it just reads plaintext ~/.netrc, that's it.

r0man:
My use case is that I want to manage all my configs with Guix and don't want any secrets in there, nor in environment variables. I read them from a single place, which is `~/.authinfo.gpg` in my case. If you want to remove `~/.authinfo(.gpg)`  in favor of `~/.netrc(.gpg)` fine with me, but removing the gpg support makes this feature useless for me ...

ericdallo:
> The emacs client is just using eca, and eca is the one needs to handle secrets. So shall we just stick to netrc in eca for standard and common practice here?

Yeah, I agree with that, even as a emacs user, it's not eca's duty to do those gpg things I guess.

Maybe since this is already implemented, we could mainly recommend .netrc and just mention about agent info as a optional way?

ericdallo:
Another idea: since authinfo is emacs specific, we could move that logic to there, spawning ECA process with token envs getting the data from authinfo, WDYT @r0man ?

r0man:
@ericdallo About moving the code: How would that work? Do you want to read the ECA config in Emacs and figure out which credentials to get via Emacs auth-source library? Sound we duplicate a lot of stuff for little benefit.

sg-qwt:
> > The emacs client is just using eca, and eca is the one needs to handle secrets. So shall we just stick to netrc in eca for standard and common practice here?
> 
> Yeah, I agree with that, even as a emacs user, it's not eca's duty to do those gpg things I guess.
> 
> Maybe since this is already implemented, we could mainly recommend .netrc and just mention about agent info as a optional way?

Since we're not stable yet, I'd suggest to revert or remedy on top of this PR with a clean/simple plaintext .netrc solution.

Secret management is definitely not the topic of ECA and extreme delicate and complex which we'd avoid as scope I'd say. 

For example, do we support password protected gpg? Or do we even support smartcard protected gpg?
Next day, comes a guy say let's support password protected age.
Then, next day, some guy say let's support AWS KMS/hashicorp vault/1password, etc..

@r0man It's user side's duty to get/decode secret from gpg/age/vault whatever and populate the plain-text standard .netrc . So put that kind of logic in eca to fit your usecase is improper imo.

r0man:
@sg-qwt I think an agent should have some way to get credentials for their user in a safe way. Ideally even from multiple sources. Emacs's auth source library does this for example. So you could also argue to remove it from there. Anyway, if you feel this is causing issues or problems and can't live with it feel free to remove it. I will find another way then.

sg-qwt:
> My use case is that I want to manage all my configs with Guix and don't want any secrets in there, nor in environment variables. I read them from a single place, which is `~/.authinfo.gpg` in my case. If you want to remove `~/.authinfo(.gpg)` in favor of `~/.netrc(.gpg)` fine with me, but removing the gpg support makes this feature useless for me ...

As a side note, although I'm not familiar with Guix. But in NixOS, we have solutions [like sops-nix](https://github.com/Mic92/sops-nix) or [agenix](https://github.com/ryantm/agenix) to do `atomic, declarative, and reproducible secret provisioning`. I'm sure Guix in that regards can have something similar (maybe already have).

But again, that's your guix logic code instead of eca here.

ericdallo:
Yeah, I can understand both points and after reading a little bit more about both standards I'm leaning to support netrc only.

@r0man ECA reads LLMs auth from configs and env vars, clients can both pass env vars to the process or custom client configs when calling initialize, the env may be the best here, where we follow a standard of `<provider>_API_KEY`.

PR #147, opened 2025-10-07
Added some tests for stopping concurrent tool calls.

bombaywalla:
Added one test for concurrent tool calls.
And another for stopping tool calls while some are running and others have completed.

Mostly copied from what was there already.
So, might be worth checking closely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Added tests for stopping concurrent tool calls. #147
+
 ## 0.65.0
 
 - Added ability to cancel tool calls. Only the shell tool currently. #145
--- a/test/eca/features/chat_test.clj
+++ b/test/eca/features/chat_test.clj
@@ -21,6 +21,18 @@
     (is (match? {:chat-id string? :status :prompting} resp))
     {:chat-id chat-id}))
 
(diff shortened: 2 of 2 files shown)


ericdallo (approved):
Looks good!

ericdallo:
@bombaywalla can you check[ flaky tests](https://github.com/editor-code-assistant/eca/actions/runs/18384392343/job/52378978302) related to those tests please?

bombaywalla:
I'll take a look.

Given that the tests rely on sleeps to ensure sequencing, I'm not too surprised that the tests are somewhat flaky.
I'll see if I can use promises to ensure the sequencing between the tool calls and the prompt stop.

ericdallo:
> Given that the tests rely on sleeps to ensure sequencing, I'm not too surprised that the tests are somewhat flaky.

I thought the same, I'm ok increasing the sleep time to a bit more (1-2s) if that will make tests 99% not flaky, but feel free to think about

bombaywalla:
Okay.  I have a fix using promises.
I did some testing myself, but maybe we just see if it is still flaky in CI.

Since this PR is already merged, I'm not sure of the etiquette involved.
Do I make another PR?
[[Added later]] I made another PR #150

PR #148, opened 2025-10-08
chore: update clojure to 1.12.2 and update flake lock

sunng87:


ericdallo (approved):
thank you!

PR #150, opened 2025-10-09
Improved flaky test

bombaywalla:

### DIFF
--- a/test/eca/features/chat_test.clj
+++ b/test/eca/features/chat_test.clj
@@ -258,15 +258,16 @@
             :call-tool-mock
             ;; Ensure that the tools complete in the 3-2-1 order by adjusting sleep times
             (fn [name & _others]
+              ;; When this is called, we are already in a future.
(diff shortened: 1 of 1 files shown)


bombaywalla:
The failing tests seems unrelated to this PR.
Would you please check?

ericdallo:
@bombaywalla am I wrong or the macos one froze on your test?

bombaywalla:
Looks that way.
If I understand the logs correctly, it is hanging on `concurrent-tool-calls-test`.
But I cannot understand why that test would hang.
There is no `deref` in that test.

bombaywalla:
I did some defensive programming and added timeouts to the derefs so that the tests should not hang.
I adjusted some of the timeouts to make the tests more likely to succeed.
I ran each of the two new tests 10 times manually and they worked.

The failing tests in the PR seem to be unrelated to my changes.

Hope this works better.

PR #165, opened 2025-10-22
Add :config-file cli option

sg-qwt:
Add :config-file cli option

### DIFF
--- a/src/eca/config.clj
+++ b/src/eca/config.clj
@@ -25,13 +25,13 @@
 (def ^:private logger-tag "[CONFIG]")
 
 (def ^:dynamic *env-var-config-error* false)
-(def ^:dynamic *cli-file-config-error* false)
+(def ^:dynamic *custom-config-error* false)
 (def ^:dynamic *global-config-error* false)
 (def ^:dynamic *local-config-error* false)
 
 (def ^:private listen-idle-ms 3000)
 
-(def config-file-path-from-cli* (atom nil))
+(def custom-config-file-path* (atom nil))
 
 (def initial-config
   {:providers {"openai" {:api "openai-responses"
@@ -153,13 +153,13 @@
 
 (def ^:private config-from-envvar (memoize config-from-envvar*))
 
-(defn ^:private config-from-cli* []
-  (when-some [path @config-file-path-from-cli*]
+(defn ^:private config-from-custom* []
+  (when-some [path @custom-config-file-path*]
     (let [config-file (io/file path)]
       (when (.exists config-file)
-        (safe-read-json-string (slurp config-file) (var *cli-file-config-error*))))))
+        (safe-read-json-string (slurp config-file) (var *custom-config-error*))))))
 
-(def ^:private config-from-cli (memoize config-from-cli*))
+(def ^:private config-from-custom (memoize config-from-custom*))
 
 (defn global-config-dir ^File []
   (let [xdg-config-home (or (get-env "XDG_CONFIG_HOME")
@@ -275,9 +275,9 @@
     [:behavior :ANY :toolCall :approval :deny :ANY :argsMatchers]
     [:otlp]]})
 
-(defn ^:private config-from-cli-or-default-location [pure-config? db]
-  (if-some [config-from-cli (config-from-cli)]
-    (when-not pure-config? config-from-cli)
+(defn ^:private config-from-custom-or-default-location [pure-config? db]
+  (if-some [config-from-custom (config-from-custom)]
+    (when-not pure-config? config-from-custom)
     (deep-merge
      (when-not pure-config? (config-from-global-file))
      (when-not pure-config? (config-from-local-file (:workspace-folders db))))))
@@ -290,7 +290,7 @@
                  eca-config-normalization-rules
                  (deep-merge initialization-config
                              (when-not pure-config? (config-from-envvar))
-                             (config-from-cli-or-default-location pure-config? db))))))
+                             (config-from-custom-or-default-location pure-config? db))))))
 
 (defn validation-error []
   (cond
--- a/src/eca/main.clj
+++ b/src/eca/main.clj
@@ -97,7 +97,7 @@
   (proxy/load!)
   (when (= "server" action)
     (when-some [cfg-file (:config-file options)]
-      (reset! config/config-file-path-from-cli* cfg-file))
+      (reset! config/custom-config-file-path* cfg-file))
     (alter-var-root #'logger/*level* (constantly (keyword (:log-level options))))
     (let [finished @(server/run-io-server! (:verbose options))]
       {:result-code (if (= :done finished) 0 1)})))


src/eca/config.clj
```diff
@@ -4,7 +4,11 @@
   1. base: fixed config var `eca.config/initial-config`.
   2. env var: searching for a `ECA_CONFIG` env var which should contains a valid json config.
   3. local config-file: searching from a local `.eca/config.json` file.
-  4. `initializatonOptions` sent in `initialize` request."
+  4. `initializatonOptions` sent in `initialize` request.
+
+  When a `ECA_CONFIGFILE` env var set, it takes precedence and we only deep merge:
+  1. base: fixed config var `eca.config/initial-config`.
+  2. env file: searching for a `ECA_CONFIGFILE` env var which points to a json config file."
```
ericdallo: I think we should just mention the env var in the existing waterfall, something like, maybe just add between 2 and 3?

src/eca/config.clj
```diff
@@ -265,17 +278,24 @@
...
   (let [initialization-config @initialization-config*
         pure-config? (:pureConfig initialization-config)]
-    (deep-merge initial-config
-                (normalize-fields
-                 eca-config-normalization-rules
-                 (deep-merge initialization-config
-                             (when-not pure-config? (config-from-envvar))
-                             (when-not pure-config? (config-from-global-file))
-                             (when-not pure-config? (config-from-local-file (:workspace-folders db))))))))
+    (if-some [envfile-config (config-from-envfile)]
```
ericdallo: I believe we can support both config env and config file env, we don't need this if, probably onone will set both ways so no need for that

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add `ECA_CONFIGFILE` env var to allow point config file explicitly.
```
ericdallo: Can we call it `ECA_CONFIG_FILE` please?

ericdallo:
@sg-qwt the feature looks good, just small changes requested

src/eca/config.clj
sg-qwt: My original imp was to snug that into waterfall between 3/4. But then I realized that didn't make sense. When user supplied explicit config path to the config file, I don't think he'd (at least myself this case) expect that to be merged with local/home config in the waterfall.

sg-qwt: For example, I generated some eca config using home-manager. Then I go
`ECA_CONFIGFILE=/nix/store/xxx/my.json eca server`
That my.json is my exact pure config to start the server, I wouldn't expect it to be further merged with other stuff this case..

ericdallo: I see your case, but:
- it's just easier to understand the waterfall with dumb logics like merging all posibilities, it's way easier to explain to users how that works
- I don't see why limit to not supporting passing 2 envs and merging them

It might be uncommon, but not impossible, if that makes code even easier to understand/maintain, I'd take that

src/eca/config.clj
sg-qwt: From your imp of ECA_CONFIG, I guess the usecase is quiet different? In your case, ECA_CONFIG is used as a convenience for user to hack a part of the config? My usecase with ECA_CONFIG_FILE as mentioned above is actually just to explicitly point config file to specified location.

src/eca/config.clj
sg-qwt: image
I can add 2 into the below case, so that when ECA_CONFIG and ECA_CONFIG_FILE both set, they get merged?


But the point is, when there is a option to set config explicitly, people usually expect it to be a replace logic instead of  going down waterfall with local config.

For example, `bb --config xxxx`
`--config <file>   Replace bb.edn with file. Defaults to bb.edn adjacent to invoked file or bb.edn in current dir. Relative paths are resolved relative to bb.edn.`

That means I want a explicit config to run with bb, not using my home config or what other config.

https://github.com/search?q=MANGOHUD_CONFIGFILE+NOT+is%3Afork+language%3ANix+&type=code
People set the config via that var to replace, not to go through the waterfall.

Or maybe we shall move ECA_CONFIG outside of the waterfall on top like this?
image

What do you think? This might be the best I think.

sg-qwt: Well, a slight mod. I think initializationOptions shall always go last on both cases, so it shall be something like this:

The top is the normal local case go through local config->home config chain, the bottom is the explicit set case:
image

ericdallo: I'd still stick with a simple waterfall merge, config waterfalls not overwriting other configs is common, if we want to overwrite we should support in other way like cljfmt with `extra-indents` option not overriidng `indents` and so on.

src/eca/config.clj
ericdallo: Hum, I see! so I think it's not a env what we want, but a `--config-file` option to CLI, does that makse sense?

src/eca/config.clj
ericdallo: Check [this comment](https://github.com/editor-code-assistant/eca/pull/165#discussion_r2452916119) I may have misunderstood the feature

src/eca/config.clj
sg-qwt: > Hum, I see! so I think it's not a env what we want, but a `--config-file` option to CLI, does that makse sense?

Yes. Weather --config-file cli option or a ECA_CONFIG_FILE env var option is the same here. Some program support both, some using env var, some using --config-file. But same functionality.

sg-qwt: If it makes it clearer, we can add a --config-file option to CLI instead of `ECA_CONFIG_FILE`

ericdallo: yeah, IMO it's way more easier to understand what will do, let's do that please

ericdallo: This way, we have all waterfall working the same, the difference is how we are going to check the configs files.
Also, it's not clear if we are overriding the global or local config, WDYT?

sg-qwt: Logic shall be the same:

if no --config-file, same old waterfall of ways to get eca config, deep merging from top to bottom:

  1. base: fixed config var `eca.config/initial-config`.
  2. env var: searching for a `ECA_CONFIG` env var which should contains a valid json config.
  3. local config-file: searching from a local `.eca/config.json` file.
  4. `initializatonOptions` sent in `initialize` request.

if --config-file = file, then only merge base with file, nothing else? (The whole point of `eca --config-file=myfile.json` is not using gloabl or local config, but myfile.json)

ericdallo: hum, I see, not sure it's good to block user from having a global config, would make sense a `--local-config-file` and a `--global-config-file`?

sg-qwt: My understanding is that --config-file just a replacement path to the (local config -> global config) path. So even with the --config-file, it can fit into your original waterfall like this?

image

sg-qwt: Well, this, technically:

image

ericdallo: Yes, your suggestion replaces both local and global with the specified path, what I wonder is: imagine a case where there are eca configs like clojureBehavior for specifed projects, so in a local config, but you wanna pass your custom config file of your preferences, so I'd expect to merge both local project config with the custom --config-file you know?
I'm starting to think that maybe we want a `--extra-config-file` (or just keep `--config-file`) which will be just a new check in the waterfall before initializationOptions, WDYT?

sg-qwt: > Yes, your suggestion replaces both local and global with the specified path, what I wonder is: imagine a case where there are eca configs like clojureBehavior for specifed projects, so in a local config, but you wanna pass your custom config file of your preferences, so I'd expect to merge both local project config with the custom --config-file you know? I'm starting to think that maybe we want a `--extra-config-file` (or just keep `--config-file`) which will be just a new check in the waterfall before initializationOptions, WDYT?

No, when user set `--config-file=something.json`, I dont expect any merge with local or global, if I find clojureBehavior useful, I would add clojureBehavior into my something.json myself. That's what I meant by explicit setting and replace.

You can take a look at babashka as well. If you run `bb --config dev.edn release` `bb --config prod.edn release` for example, do I expect dev.edn merge with my project local bb.edn or prod.edn merge with my project local bb.edn? No. I expect explicit config as is specified. The bb.edn is no longer relevant.

ericdallo: I'd expect to merge them hehe, I think this is just a decision to take, there is no real standard TBH.
Well, I think it's ok to understand config-file will be used in place of global + local, would be nice though to have a more clear name that would understand it's related to a override, but I'm ok with that for now

sg-qwt:
@ericdallo pls check the new imp with cli option

src/eca/config.clj
```diff
@@ -23,11 +25,14 @@
 (def ^:private logger-tag "[CONFIG]")
 
 (def ^:dynamic *env-var-config-error* false)
+(def ^:dynamic *cli-file-config-error* false)
 (def ^:dynamic *global-config-error* false)
 (def ^:dynamic *local-config-error* false)
 
 (def ^:private listen-idle-ms 3000)
 
+(def config-file-path-from-cli* (atom nil))
```
ericdallo: can we not couple with cli here? just call it `custom-config-file`

ericdallo: same in other places of this file where we mention cli

ericdallo: In the future eca may have other ways besides CLI, that's why I'd like to not couple those names

ericdallo (commented):
Looks good! just a small thing about naming

src/eca/config.clj
sg-qwt: Done

sg-qwt:
Fixed the naming.

src/eca/main.clj
```diff
@@ -42,7 +42,7 @@
 (def log-levels #{"error" "warn" "info" "debug"})
 
 (def cli-spec
-  {:order [:help :version :verbose]
+  {:order [:help :version :verbose :config-file]
```
sg-qwt, line 45: @ericdallo I noticed you left out --log-level in cli print, is that by design?

ericdallo: I think I missed that!

ericdallo (approved):
Looks great! thank you!

PR #176, opened 2025-10-27
Improve single-occurrence file edits

zikajk:
- Add smart-edit (exact→normalized→flexible→regex) for single-occurrence; enforce uniqueness and return :strategy.
- Quote regex replacements; preserve indentation, line endings, and trailing newline.
- Route eca_edit_file: all_occurrences=true uses literal exact/normalized; default uses smart-edit; align preview.
- Update docs and add focused unit tests for smart-edit and text-match.


ericdallo:
I fixed flaky integration tests on master (besides macos sometimes failing)

PR #179, opened 2025-10-27
Simplify credential reading, support plain-text netrc only.

sg-qwt:
This is a follow up on https://github.com/editor-code-assistant/eca/pull/146 to simplify our credential reading logic.

1. drop authinfo and gpg decryption logic out of scope
2. added a --netrc-file option for flexible use-cases

Basically same as curl's implementation.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,17 +25,13 @@ Code Style
 - Unit tests that use file paths and uris should rely on `h/file-path` and `h/file-uri` to avoid windows issues with slashes.
 
 Secrets Management
-- `src/eca/secrets/netrc.clj` - Netrc format parser (multi-line format: machine/login/password/port keywords)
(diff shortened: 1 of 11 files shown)


sg-qwt:
> I was meaning to ask to add tests to the --netrc-file, but then I started to wonder: what is the best here: a `--netrc-file` CLI option or a ECA config `:netrc-file`.

I think --netrc-file in cli would be most convenient in case people would like to improvise in shell (pipe things around etc..).

A ECA config `:netrc-file` shall also work, and makes the cli a little cleaner. Even a env var like ECA_NETRC_FILE would be fine.

Personally I don't have a preference. Let me know what you think.

Also what's the best way to test this? An integration test?

ericdallo:
I prefer to have netrc-file as a config for now, probably will fix all cases, people can always have local configs for that.

having config, you will only need to test the netrc feature itself reading from config as unit tests

sg-qwt:
@ericdallo pls check the new imp with netrcFile ECA config

PR #182, opened 2025-10-29
Switch context partials to lines-range

zikajk:
- Replace :partial with :lines-range in refined contexts
- Update contexts header copy and file tag rendering
- Adjust prompt/context tests to new format


zikajk:
@ericdallo Tests are green locally, I guess it is some flakiness.

ericdallo:
@zikajk the windows test will not be the same if you run on a Unix that why.
Windows handle paths and line breaks differently, we have helpers to handle those I'm tests but you need to use, I can take a look soon

ericdallo:
I will fix the windows test on master

PR #189, opened 2025-11-01
Make it more clear how to use /login command and provide clear instructions

lukaszkorecki:
I've run into this small UX issue this morning - typing `/login` shows providers list, but doesn't really explain how to authenticate. I tried `/login github-copilot` and nothing happened 🙃 
I realized that I have to type the provider name as the next step after looking at eca.dev docs. 

This change provides a better feedback loop for invalid inputs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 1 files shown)


ericdallo (approved):

ericdallo:
Thank you!

PR #216, opened 2025-11-24
Add new supported github-copilot models to builtin conf

zikajk:


zikajk:
@ericdallo There is a problem with Gemini responses that should be fixed first:
```
'https://api.githubcopilot.com/chat/completions'
[OPENAI-CHAT] 4813 Unexpected response status: %s body: %s 400 {"error":{"message":"invalid request body","code":"invalid_request_body"}}
```

ericdallo:
Hum, we need to understand if we need to change the body for that model, maybe do in other PR the gemini?

zikajk:
Agree, done :-)

PR #225, opened 2025-11-29
Advanced hooks

zikajk:
Implement advanced lifecycle hooks

Key changes:
- New Hook Types: `sessionStart`, `sessionEnd`, `chatStart`, `chatEnd` for better lifecycle management.
- Enhanced `preRequest`: Supports prompt rewriting, context injection, and request blocking.
- Enhanced `preToolCall`: Allows argument modification, approval overrides (allow/ask/deny), and tool blocking via exit codes or JSON.
- New `postToolCall`: Provides access to tool responses and errors, allowing for post-execution context injection.
- JSON Protocol: Hooks now communicate via structured JSON on stdin/stdout for robust data exchange.
- Integration: Updates chat loop and tool state machine to respect hook decisions.

### DIFF
--- a/src/eca/features/prompt.clj
+++ b/src/eca/features/prompt.clj
@@ -35,7 +35,7 @@
 (defn ^:private replace-vars [s vars]
   (reduce
    (fn [p [k v]]
-     (string/replace p (str "{" (name k) "}") (str v)))
+     (string/replace p (str "{{" (name k) "}}") (str v)))
    s
    vars))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Enhanced hooks documentation with new types (sessionStart, sessionEnd, chatStart, chatEnd), JSON input/output schemas, execution options (timeout)
 
 - Fix custom tools to support argument numbers.
+- Improve read_file summary to mention offset being read.
 
 ## 0.84.2
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -267,6 +267,8 @@ Placeholders in the format `{{argument_name}}` within the `command` string will 
 You can configure custom command prompts for project, global or via `commands` config pointing to the path of the commands.
 Prompts can use variables like `$ARGUMENTS`, `$ARG1`, `ARG2`, to replace in the prompt during command call.
 
+You can configure in multiple different ways:
+
 === "Local custom commands"
 
     A `.eca/commands` folder from the workspace root containing `.md` files with the custom prompt.
@@ -274,6 +276,8 @@ Prompts can use variables like `$ARGUMENTS`, `$ARG1`, `ARG2`, to replace in the 
     ```markdown title=".eca/commands/check-performance.md"
     Check for performance issues in $ARG1 and optimize if needed.
     ```
+    
+    ECA will make available a `/check-performance` command after creating that file.
 
 === "Global custom commands"
 
@@ -282,6 +286,8 @@ Prompts can use variables like `$ARGUMENTS`, `$ARG1`, `ARG2`, to replace in the 
     ```markdown title="~/.config/eca/commands/check-performance.md"
     Check for performance issues in $ARG1 and optimize if needed.
     ```
+    
+    ECA will make available a `/check-performance` command after creating that file.
 
 === "Config"
 
@@ -292,6 +298,8 @@ Prompts can use variables like `$ARGUMENTS`, `$ARG1`, `ARG2`, to replace in the 
       "commands": [{"path": "my-custom-prompt.md"}]
     }
     ```
+    
+    ECA will make available a `/my-custom-prompt` command after creating that file.
 
 ## Rules
 
--- a/resources/prompts/additional_system_info.md
+++ b/resources/prompts/additional_system_info.md
@@ -1,6 +1,6 @@
 ## Project Environment Context
 
-Workspaces: {workspaceRoots}
+Workspaces: {{workspaceRoots}}
 
 **Path Resolution & Context:**
 *Workspaces:* Directories containing code or data relevant to this session.
--- a/src/eca/features/tools.clj
+++ b/src/eca/features/tools.clj
@@ -3,6 +3,7 @@
    eca native tools and MCP servers."
   (:require
    [clojure.string :as string]
+   [clojure.walk :as walk]
    [eca.features.tools.chat :as f.tools.chat]
    [eca.features.tools.custom :as f.tools.custom]
    [eca.features.tools.editor :as f.tools.editor]
@@ -38,14 +39,29 @@
     (fn [tool]
       (assoc-some tool :disabled (contains? disabled-tools (:name tool))))))
 
+(defn ^:private replace-string-values-with-vars
+  "walk through config parsing dynamic string contents if value is a string."
+  [m vars]
+  (walk/postwalk
+   (fn [x]
+     (if (string? x)
+       (reduce
+        (fn [s [k v]]
+          (string/replace s (str "{{" (name k) "}}") (str v)))
+        x
+        vars)
+       x))
+   m))
+
 (defn ^:private native-definitions [db config]
   (into
    {}
    (map (fn [[name tool]]
           [name (-> tool
                     (assoc :name name)
-                    (update :description #(-> %
-                                              (string/replace #"\{workspaceRoots\}" (constantly (tools.util/workspace-roots-strs db))))))]))
+                    (replace-string-values-with-vars
+                      {:workspaceRoots (tools.util/workspace-roots-strs db)
+                       :readFileMaxLines (get-in config [:toolCall :readFile :maxLines])}))]))
    (merge {}
           f.tools.filesystem/definitions
           f.tools.shell/definitions
--- a/src/eca/features/tools/filesystem.clj
+++ b/src/eca/features/tools/filesystem.clj
@@ -70,7 +70,7 @@
                                                        ["path" (complement fs/directory?) "$path is a directory, not a file"]]))
       (let [line-offset (or (get arguments "line_offset") 0)
             limit (->> [(get arguments "limit")
-                        (get-in config [:toolCall :readFile :maxLines] 2000)]
+                        (get-in config [:toolCall :readFile :maxLines])]
                        (filter number?)
                        (apply min))
             full-content-lines (string/split-lines (slurp (fs/file (fs/canonicalize (get arguments "path")))))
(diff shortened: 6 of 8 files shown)


zikajk:
@ericdallo I am not sure about the failing Windows test. Maybe you have an idea?

docs/examples.md
```diff
@@ -5,6 +5,60 @@ ECA config examples showing the power of its features and flexibility
 If you think your config is relevant to be shared for other people, [open a pull request here](https://github.com/editor-code-assistant/eca/edit/master/docs/examples.md)
 
 ## From users
+??? info "Hook: fix unbalanced CLJ parens (@zikajk)"
```
ericdallo, line 8: 💯

src/eca/features/prompt.clj
```diff
@@ -105,7 +108,7 @@
    (when (seq refined-contexts)
      ["## Contexts"
       ""
-      (contexts-str refined-contexts repo-map*)])
+      (contexts-str refined-contexts repo-map* (get-in db [:chats chat-id :startup-context]))])
```
ericdallo, line 111: I think this should go indeed to a refined-context, following others

src/eca/features/tools/shell.clj
```diff
@@ -15,6 +15,40 @@
...
+  {:pre [(some? cwd)
+         (or (some? script) (some? file))
+         (not (and script file))]}
+  (let [win? (string/starts-with? (System/getProperty "os.name") "Windows")
+        cmd (cond
+              (and win? file)
+              ["powershell.exe" "-ExecutionPolicy" "Bypass" "-File" file]
+
+              (and win? script)
+              ["powershell.exe" "-NoProfile" "-Command" script]
```
ericdallo, line 38: nice

src/eca/features/hooks.clj
```diff
@@ -1,76 +1,177 @@
...
+  "Attempts to parse hook output as JSON. Returns parsed map if successful, nil otherwise."
+  [output]
+  (when (and output (not-empty output))
+    (try
+      (let [parsed (json/parse-string output true)]
+        (if (map? parsed)
+          parsed
+          (logger/debug logger-tag "Hook JSON output must result in map")))
+      (catch Exception e
+        (logger/debug logger-tag "Hook output is not valid JSON, treating as plain text"
```
ericdallo, line 42: I believe these should be at least a logger/warn?

src/eca/features/hooks.clj
```diff
@@ -1,76 +1,177 @@
...
+
+(defn ^:private should-skip-on-error?
+  "Check if postToolCall hook should be skipped when tool errors.
+  By default, postToolCall hooks only run on success unless runOnError is true."
+  [type hook data]
+  (and (= type :postToolCall)
+       (not (get hook :runOnError false))
+       (:error data)))
+
+(defn hook-matches? [hook-type data hook]
```
ericdallo:
```suggestion
(defn ^:private hook-matches? [hook-type data hook]
```

src/eca/features/chat.clj
```diff
@@ -46,22 +46,57 @@
...
+          content))
+
+(defn ^:private append-post-tool-additional-context!
+  "Append additionalContext (wrapped as XML) from a postToolCall hook to the
+   matching tool_call_output message so LLM sees it in the next round."
+  [db* chat-id tool-call-id hook-name additional-context]
+  (when (not (string/blank? additional-context))
+    (let [entry {:type :text :text (wrap-additional-context hook-name additional-context)}]
+      (swap! db* update-in [:chats chat-id :messages]
+             ;; Optimized: Scans messages backwards since the tool output is likely one of the last items.
```
ericdallo, line 83: this is the kind of optmization I don't think we will ever need, especially because messages may have < 1000 elements most of the time? I doubt it will be a problem, and makes harder to understand the code, but not a big deal too

src/eca/features/chat.clj
```diff
@@ -532,362 +704,437 @@
...
+                                          (assert-chat-not-stopped! chat-ctx)
+                                          (let [tool (tool-by-full-name full-name all-tools)]
+                                            (transition-tool-call! db* chat-ctx id :tool-prepare
+                                                                   {:name           (:name tool)
+                                                                    :server         (:name (:server tool))
+                                                                    :full-name      full-name
+                                                                    :origin         (:origin tool)
+                                                                    :arguments-text arguments-text
+                                                                    :summary        (f.tools/tool-call-summary all-tools full-name nil config)})))
+            :on-tools-called            (on-tools-called! chat-ctx received-msgs* add-to-history!)
```
ericdallo: I don't really like this vertical alignment in formatting in the changes of this file, especially because I use only a notebook so it makes harder to see the code and I need to keep going to right and left, can we rollback?

src/eca/features/chat.clj
```diff
@@ -532,362 +704,437 @@
...
+                  (when-not (#{:stopping :cleanup :rejected} (:status (get-tool-call-state @db* chat-id id)))
+                    (case decision
+                      :ask   (transition-tool-call! db* chat-ctx id :approval-ask {:progress-text "Waiting for tool call approval"})
+                      :allow (transition-tool-call! db* chat-ctx id :approval-allow {:reason reason})
+                      :deny  (transition-tool-call! db* chat-ctx id :approval-deny {:reason reason})
+                      (logger/warn logger-tag "Unknown value of approval" {:approval decision :tool-call-id id})))
+                  (if (and @approved?* (not hook-rejected?))
+                    (when-not (#{:stopping :cleanup} (:status (get-tool-call-state @db* chat-id id)))
+                      (assert-chat-not-stopped! chat-ctx)
+                      (let [delayed-future (delay (future (let [result               (f.tools/call-tool! name
```
ericdallo: A few enters + remove vertical alignment would make this way better to read

docs/examples.md
```diff
@@ -5,6 +5,60 @@ ECA config examples showing the power of its features and flexibility
...
+
+	First install latest [[babashka](https://github.com/babashka/babashka)] + [[bbin](https://github.com/babashka/bbin)].
+	Then run:
+	```bbin install https://github.com/bhauman/clojure-mcp-light.git --as clj-paren-repair --main-opts '["-m"  "clojure-mcp-light.paren-repair"]'```
+
+    ```javascript title="config.json"
+	{...
+	 "hooks:{"CLJ-balanced-parens-check": {"type":"postToolCall",
```
ericdallo:
```suggestion
	 "hooks": {"CLJ-balanced-parens-check": {"type":"postToolCall",
```

docs/examples.md
```diff
@@ -5,6 +5,60 @@ ECA config examples showing the power of its features and flexibility
...
+                                           "matcher": "eca__write_file|eca__edit_file",
+                            		       "actions": [{"type": "shell",
+                  					                    "file": "/Users/akiz/.config/eca/hooks/clj_check_parens.sh"}]}}
```
ericdallo: Relative may be more friendly
```suggestion
                  					                    "file": "hooks/clj_check_parens.sh"}]}}
```

docs/examples.md
```diff
@@ -5,6 +5,60 @@ ECA config examples showing the power of its features and flexibility
...
+	...}
+	```
+
+	```bash title="/Users/akiz/.config/eca/hooks/clj_check_parens.sh"
```
ericdallo:
```suggestion
	```bash title="~/.config/eca/hooks/clj_check_parens.sh"
```

integration-test/integration/chat/hooks_test.clj
```diff
@@ -0,0 +1,382 @@
...
+                "sessionEnd"]
+               lines))))))
+
+(deftest posttoolcall-receives-tool-response-test
+  (testing "postToolCall hook receives tool_response and tool_input after tool execution"
+    (eca/start-process!)
+
+    (llm.mocks/set-case! :tool-calling-0)
+
+    (let [log-path (str h/default-root-project-path "/.eca/posttool-log.txt")]
```
ericdallo: Almost sure this is causing windows integration tests to fail, we should not concat paths with str, but use io/file or fs/file

src/eca/features/prompt.clj
zikajk: It does, right?

src/eca/features/hooks.clj
zikajk: I am not sure. It is still valid to not return JSON. The question would be if this should be logged at all :-).

src/eca/features/chat.clj
zikajk: The code is a little more complex, but I believe it's fine with the comment. I agree that a map with over 1000 elements is relatively okay at this moment, but for the future, is this perhaps safer? Or does it really strike you as over-optimization?

src/eca/features/chat.clj
zikajk: Yeah, I agree. 
On my other projects this is must-to-have, so I have it in my global AGENTS.md but I am not big fan neither :-). 
Updated

src/eca/features/chat.clj
zikajk: I removed the vertical alignment, but where do you see newlines would help?

docs/examples.md
zikajk: done

src/eca/features/hooks.clj
zikajk: Done

docs/examples.md
zikajk: Done

src/eca/features/hooks.clj
ericdallo: yeah maybe the first one? I don't know too

src/eca/features/chat.clj
ericdallo: No, I think it's good for now because of the coment + only few lines more complex

src/eca/features/chat.clj
ericdallo: I think one after delay and other after future maybe?

src/eca/features/hooks.clj
zikajk: I think both are valid. 
The first one will tell the user that his output is a valid json but not the one compatible with advanced hooks. 
The second one might be useful to catch if the JSON is invalid but it will probably throw even with a string. 
If i am not wrong, they will fire too often to have them on any other level than `debug`, no?

src/eca/features/chat.clj
zikajk: Ah, i see, i did that :). Good.

PR #232, opened 2025-12-13
fix #231 - Add version parameter to openai-chat provider config to support different API versions

### LINKED ISSUE #231: Mistral AI API tool call issues.
**Describe the bug**
Just running a test "this is a test" against Mistral AI API is giving this error-

```
LLM response status: 422 body: {"object":"error","message":{"detail":[{"type":"extra_forbidden","loc":["body","max_completion_tokens"],"msg":"Extra inputs are not permitted","input":32000}]},"type":"invalid_request_error","param":null,"code":null}
```

Mistral AI config

```json
"mistral": {
      "api": "openai-chat",
      "url": "https://api.mistral.ai/",
      "completionUrlRelativePath": "/v1/chat/completions",
      "key": "XXXXXXXXX",
      "models": {
        "labs-devstral-small-2512": {}
      }
    }
```

look like it is a know issue caused when `max_completion_tokens` property is sent.

can I remove that?

Tested API with GPTEL and Curl and it works fine.

```
(gptel-make-openai "MistralLeChat" 
      :host "api.mistral.ai"
      :endpoint "/v1/chat/completions"
      :protocol "https"
      :key "XXXXXXXXXX"
      :models '("labs-devstral-small-2512"))
```

CsBigDataHub:
- Add version parameter to openai-chat provider config to support different API versions
- Version 1: Uses max_tokens parameter (for Mistral AI and older OpenAI-compatible endpoints)
- Version 2 (default): Uses max_completion_tokens parameter (for OpenAI and modern endpoints)
- Fixes Mistral AI 422 errors by allowing configuration of appropriate token parameter
- Supports both provider-level and model-level version configuration with provider taking precedence
- Maintains full backward compatibility by defaulting to version 2
- Automatically removes conflicting token parameters based on version
- Prevents sending version parameter to external APIs
- Includes comprehensive integration tests
- Updates documentation with examples and schema
- Adds CHANGELOG entry under Unreleased section

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+- Add `version` parameter to `openai-chat` provider config to support different API versions:
+  - **Version 1**: Uses `max_tokens` parameter (for Mistral AI and older OpenAI-compatible endpoints)
+  - **Version 2** (default): Uses `max_completion_tokens` parameter (for OpenAI and modern endpoints)
+  - Fixes Mistral AI 422 errors by allowing configuration of appropriate token parameter
+  - Supports both provider-level and model-level version configuration with provider taking precedence
+  - Maintains full backward compatibility by defaulting to version 2
+
 ## 0.86.0
 
 - Improve agent behavior prompt to mention usage of editor_diagnostics tool. #230


ChetanKoneru:
@ericdallo please do NOT merge, I am having tool call issues

```
LLM response status: 400 body: {"object":"error","message":"Tool call id was 7878-452755744 but must be a-z, A-Z, 0-9, with a length of 9.","type":"invalid_function_call","param":null,"code":"3280"}
```

ericdallo:
@CsBigDataHub let me think a little bit, I don't want to have different functions, apis or complex things just to one or other param more or removed, I want user to be able to customize that like we did for responses adn anthropid, we may manage to do the same, let me take a look

PR #234, opened 2025-12-13
feat: support Google Gemini thought signatures + gemini 3 support

ArthurHeymans:

feat: support Google Gemini thought signatures

Implements support for capturing and relaying 'thought signatures'
required by Google Gemini 3 Pro for function calling. See
https://ai.google.dev/gemini-api/docs/thought-signatures "Thought
signatures are encrypted representations of the model's internal thought
process and are used to preserve reasoning context across multi-turn
interactions."

This change does the following:

- Persist thought signatures in tool call state and pass them back via
  `extra_content`.
- Add `skipThoughtSignatureValidator` config to bypass validation for custom tool calls.
- Register `gemini-3-pro-preview` model.

Co-Authored-By: eca <noreply@eca.dev>

### DIFF
--- a/integration-test/llm_mock/openai.clj
+++ b/integration-test/llm_mock/openai.clj
@@ -152,7 +152,7 @@
         (sse-send! ch "response.function_call_arguments.delta"
                    {:type "response.function_call_arguments.delta"
                     :item_id "item-1"
-                    :delta (str "h\":\"" (h/project-path->canon-path "resources") "\"}")})
+                    :delta (str "h\":\"" (h/json-escape-path (h/project-path->canon-path "resources")) "\"}")})
         ;; Complete with the function call present so the client triggers tools
         (sse-send! ch "response.completed"
                    {:type "response.completed"
--- a/integration-test/llm_mock/openai_chat.clj
+++ b/integration-test/llm_mock/openai_chat.clj
@@ -114,7 +114,7 @@
   (send-sse! ch {:choices [{:delta {:tool_calls [{:index 0
                                                    :function {:arguments "{\"pat"}}]}}]})
   (send-sse! ch {:choices [{:delta {:tool_calls [{:index 0
-                                                   :function {:arguments (str "h\":\"" path "\"}")}}]}}]})
+                                                   :function {:arguments (str "h\":\"" (h/json-escape-path path) "\"}")}}]}}]})
   (send-sse! ch {:usage {:prompt_tokens 5 :completion_tokens 30}})
   (send-sse! ch {:choices [{:delta {} :finish_reason "tool_calls"}]})
   (hk/close ch))


ericdallo:
@ArthurHeymans Thank you for helping with that!
I beleive this is the same openai and anthropic use [here](https://github.com/editor-code-assistant/eca/blob/eb122c0f6eae5a6397bc9db6c95ef6e723f26e7e/src/eca/llm_providers/openai.clj#L175) and [here](https://github.com/ericdallo/eca/blob/eb122c0f6eae5a6397bc9db6c95ef6e723f26e7e/src/eca/llm_providers/anthropic.clj#L224) right?
I know external-id is not the best name, but it's internal and we can consider renaming it in a next PR, but for this one, can we use the same standard of those other providers? returning the external-id for those reasons and including in next requests like done in others?

Also, what's the rationale behind a flag to skip sending those? can't we just send if they are there?

ArthurHeymans:
> @ArthurHeymans Thank you for helping with that! I beleive this is the same openai and anthropic use [here](https://github.com/editor-code-assistant/eca/blob/eb122c0f6eae5a6397bc9db6c95ef6e723f26e7e/src/eca/llm_providers/openai.clj#L175) and [here](https://github.com/ericdallo/eca/blob/eb122c0f6eae5a6397bc9db6c95ef6e723f26e7e/src/eca/llm_providers/anthropic.clj#L224) right? I know external-id is not the best name, but it's internal and we can consider renaming it in a next PR, but for this one, can we use the same standard of those other providers? returning the external-id for those reasons and including in next requests like done in others?
> 
Ok I changed the name.

> Also, what's the rationale behind a flag to skip sending those? can't we just send if they are there?

I think it has to be used if gemini is used after another model, but it indeed makes little sense to have this in the configuration. I removed it.

ericdallo (commented):
That looks way better! May I ask to improve the [google integration-test](/home/greg/dev/eca/integration-test/integration/chat/google_test.clj) adding this case please?

You can run integration-tests with `bb integration-test`, you can comment other tests and leave only google in entrypoint.clj to quickly test

ArthurHeymans:
Accidentally pushed test before it was ready

src/eca/features/chat.clj
```diff
@@ -1133,6 +1134,7 @@
                                                               :full-name full-name
                                                               :origin (:origin tool)
                                                               :arguments-text arguments-text
+                                                              :external-id external-id
```
ericdallo: I'm trying to understand why this is required

ArthurHeymans:
Regarding the question about why  is required at line 1137:

You're right - it's NOT required here. The `external-id` is already preserved in the `tool-call` object parameter (from the LLM provider at line 828) and gets added to history correctly at lines 886 and 948 where we use the original `tool-call` object with `assoc`.

The tool-call state is only used for tracking execution status and doesn't need `external-id` since we never reconstruct tool-calls from the state for history.

I'll remove it from the `:tool-prepare` transition.

ArthurHeymans:
I've pushed a commit removing the unnecessary `:external-id` from the tool-call state. The `external-id` is still properly preserved and sent back to Google Gemini through the `tool-call` object that gets added to history.

The tests pass and the integration test should still work correctly since the external-id flow through `openai_chat.clj` is unchanged.

ericdallo:
@ArthurHeymans I believe there are things to fix on integration test that are failing for windows, probably related to slash issues

ericdallo (approved):
Looks great, thank you a lot!

MrnoRac:
I've run into a small issue with gemini 3-pro-preview where after reading files it fails with:
```json
LLM response status: 400 body: [{
  "error": {
    "code": 400,
    "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:eca__read_file` , position 7. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.",
    "status": "INVALID_ARGUMENT"
  }
}
]
```
image


I have the latest installion on emacs (nothing fancy in there other than prelude stuff I suppose), running on manjaro.

Let me know if I can do something to get information to debug this.

MrnoRac:
@ArthurHeymans (Forgot to mention in the main comment above)

PR #235, opened 2025-12-14
Fix DeepSeek reasoning support in openai-chat API

zikajk:
- Pass reasoning_content through the pipeline for DeepSeek tool loops
- Merge adjacent assistant messages only when reasoning_content present
- Add prune-history to drop stale reasoning from conversation history
- Consolidate reasoning state into single atom for cleaner state management
- Extract find-last-msg-idx utilities to llm-util for reuse
- Add comprehensive tests for DeepSeek reasoning scenarios

### DIFF
--- a/src/eca/features/chat.clj
+++ b/src/eca/features/chat.clj
@@ -1130,7 +1130,7 @@
                                                               :arguments-text arguments-text
                                                               :summary (f.tools/tool-call-summary all-tools full-name nil config)})))
             :on-tools-called (on-tools-called! chat-ctx received-msgs* add-to-history!)
-            :on-reason (fn [{:keys [status id text external-id reasoning-content]}]
+            :on-reason (fn [{:keys [status id text external-id delta-reasoning?]}]
                          (assert-chat-not-stopped! chat-ctx)
                          (case status
                            :started  (do (swap! reasonings* assoc-in [id :start-time] (System/currentTimeMillis))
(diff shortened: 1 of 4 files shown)


src/eca/features/chat.clj
```diff
@@ -1135,20 +1130,22 @@
...
                            :thinking (do (swap! reasonings* update-in [id :text] str text)
                                          (send-content! chat-ctx :assistant {:type :reasonText :id id :text text}))
-                           :finished (let [total-time-ms (- (System/currentTimeMillis) (get-in @reasonings* [id :start-time]))]
-                                       (add-to-history! {:role "reason"
-                                                         :content {:id id
-                                                                   :external-id external-id
-                                                                   :total-time-ms total-time-ms
-                                                                   :text (get-in @reasonings* [id :text])}})
-                                       (send-content! chat-ctx :assistant {:type :reasonFinished :total-time-ms total-time-ms :id id}))
+                           :finished (when-let [start-time (get-in @reasonings* [id :start-time])]
```
zikajk, line 1140: @ericdallo I changed this because when I clicked on stop in the right moment (after question and moment before reasoning started) then I got an error on `(- <time> nil)`.

src/eca/llm_providers/openai_chat.clj
```diff
@@ -345,66 +412,53 @@
...
-                                  (reset! content-buffer* "")))
-                              (when @reasoning-started*
-                                (on-reason {:status :finished :id @current-reason-id*})
-                                (reset! reasoning-started* false)
-                                (reset! reasoning-type* nil))
+                              (flush-content-buffer)
+                              (finish-reasoning! reasoning-state* on-reason)
                               (execute-accumulated-tools! tool-calls* on-tools-called-wrapper on-tools-called handle-response))
                             (when (seq (:choices data))
                               (doseq [choice (:choices data)]
```
zikajk, line 432: @ericdallo This also doesn't make a sense (we have only one choice and I decided to throw an error if we have more), but I will keep it for a next PR.

ericdallo: I think we would need to understand if any provider return more than a choice

zikajk:
@ericdallo Some improvements ;)

PR #239, opened 2025-12-18
Add git worktree support

ArthurHeymans:

Add git worktree support for workspace and config discovery

- Add new eca.git namespace with utilities for detecting git roots and worktrees
- Update config loading to fall back to git root when no workspace folders provided
- Update initialize handler to use git root as workspace folder fallback
- Fix test isolation by mocking config/get-env in provider-api-key test
- Add comprehensive tests for git utilities and config fallback behavior

Signed-off-by: Arthur Heymans <arthur@aheymans.xyz>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix deepseek reasoning with openai-chat API #228
 - Support `~` in dynamic string parser.
 - Support removing nullable values from LLM request body if the value in extraPayload is null. #232
+- Add git worktree support: automatically detect and use git repository root as workspace folder when not explicitly provided.
 
 ## 0.86.0
 
--- a/src/eca/config.clj
+++ b/src/eca/config.clj
@@ -17,6 +17,7 @@
    [clojure.string :as string]
    [clojure.walk :as walk]
    [eca.logger :as logger]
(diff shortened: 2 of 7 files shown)


zikajk:
What problem does it solve? Thank you

ArthurHeymans:
> What problem does it solve? Thank you

I have a git worktree inside a git directory, let's say .worktree/feature_x. Previously eca would then operate on the root worktree rather than the feature_x worktree.

ericdallo:
Please create issues before PRs for things we should discuss before, this avoids re-work
Not sure I agree with this one, workspace roots are client responsibilities for some reasons, because each client handle workspace roots diferently.
I wonder if one can solve that starting eca with different workspace roots like it already supports

PalaceChan:
agree with ericdallo

> I wonder if one can solve that starting eca with different workspace roots like it already supports

That is exactly what I do. I have different worktrees in different roots and they each have my own `.eca` (symlink pointing to common eca configuration) and i simply start eca in the different roots of the different worktrees. Works perfectly well that way.

ericdallo:
Closing this as this was done in a different way [here](https://github.com/editor-code-assistant/eca-emacs/issues/165)

PR #243, opened 2025-12-21
 Fix OpenAI chat tool_call id handling in streaming

zikajk:
Preserves provider-generated tool_calls[].id in OpenAI-compatible Chat Completions and stabilizes streaming tool-call accumulation by index/id, ensuring tool-call UI/state transitions remain correct across providers.


ericdallo:
Thank you!

PR #245, opened 2025-12-26
Add dynamic model discovery via /models endpoint

zikajk:
Add fetchModels option to enable automatic model discovery from OpenAI-compatible providers.

- fetchModels: true enables dynamic model discovery from /models endpoint
- Static model configs override discovered models for customization
- Graceful error handling with user-visible warnings

The feature is opt-in and only works with providers exposing OpenAI-compatible /models endpoints.

### DIFF
--- a/src/eca/llm_api.clj
+++ b/src/eca/llm_api.clj
@@ -62,6 +62,27 @@
 (defn ^:private real-model-name [model model-capabilities]
   (or (:model-name model-capabilities) model))
 
+(defn provider->api-handler [provider config]
+  (cond
+    (= "openai" provider) {:api :openai-responses
+                           :handler llm-providers.openai/create-response!}
+    (= "anthropic" provider) {:api :anthropic
+                              :handler llm-providers.anthropic/chat!}
+    (= "github-copilot" provider) {:api :openai-chat
+                                   :handler llm-providers.openai-chat/chat-completion!}
+    (= "google" provider) {:api :openai-chat
+                           :handler llm-providers.openai-chat/chat-completion!}
+    (= "ollama" provider) {:api :ollama
+                           :handler llm-providers.ollama/chat!}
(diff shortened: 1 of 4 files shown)


ericdallo (commented):
That's cool! will take a closer look soon

zikajk:
@ericdallo I thought about making this the default when no models are filled in for a provider (for better UX), but fetching models takes some time so I decided not to (yet?).  

**Motivation**: I have a lot of providers and I am lazy to fill in all models I might use. 
I want to see when MiniMax2.1 will be available on Synthetic.new ;).

src/eca/llm_api.clj
```diff
@@ -178,7 +178,7 @@
           :extra-payload extra-payload}
          callbacks)
 
-        model-config
+        (:api provider-config)
```
zikajk: @ericdallo I think this is the only place that can introduce some regression. 
I am not quite sure what was the idea behind `model-config` check.

ericdallo:
> @ericdallo I thought about making this the default when no models are filled in for a provider (for better UX), but fetching models takes some time so I decided not to (yet?).

@zikajk yes makes sense, but we already have the `models/sync-models!` which we can run async after getting them and update UI, would be possible to do that async after startup no? In another PR

src/eca/llm_api.clj
ericdallo: it's not clear, but if there is a model-config, user has some config for a provider that we don't support built-in, so it's a custom provider with a model that matches the requested model

ericdallo: with this change, we only check if the provider matches the requested model provider, but not the model name, but since we have dynamic it may be the only way, maybe we should:
- if dynamic-model is avaialble for that api, keep this check like that
- if not do the current check

WDYT?

zikajk: Yes, I think it is a better approach.

Issue #248, opened 2025-12-29
Support basic proxy authentication via PROXY_HTTP / PROXY_HTTPS


ikappaki:
I'd like to propose adding support for basic proxy authentication via PROXY_HTTP / PROXY_HTTPS (e.g., http://`username:password@proxy.server`) for environments that require authenticated proxies.

I’ll follow up with a PR containing a candidate implementation.

Thanks!

To upvote this issue, give it a thumbs up. See [this list](https://github.com/editor-code-assistant/eca/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

ericdallo:
@ikappaki is that related to https://github.com/editor-code-assistant/eca/issues/73? we added support for HTTP_PROXY and HTTPS_PROXY

ikappaki:
Hi @ericdallo,

this is regarding support for the `username:password` basic authentication scheme in `PROXY_HTTP[S]` values (e.g., `http://username:password@proxy.server:port`). Unless I’m gravely mistaken, this is not currently supported by the `http[s].proxyHost` Java property used by eca.

Thanks

ericdallo:
Got it, thank you, will take a look, a 2k PRs needs to have a careful review :)

PR #249, opened 2025-12-29
Add basic proxy authentication support via HTTP[S]_PROXY

ikappaki:
Hi,

can you please consdiser reviewing patch to add support for basic proxy authentication via the HTTP[S]_PROXY environment variables. It addresses #248.

The implementation extracts the `username` and `password` from the proxy URL and passes them to Hato by setting an authenticator option in `build-http-client` (as `{:user "username" :pass "password"}`). Unfortunately, configuring a global `java.net.Authenticator` does not work here, as the Java HTTP client used by Hato does not honor it, so the authenticator needs to be provided on each hato request instead.

Because it's critical that this option is never omitted from any hato request (otherwise some calls would skip basic authentication), I added a `eca.client-test-helpers/with-client-proxied` test macro that routes Hato requests through an embedded `LittleProxy` instance. This makes it possible to reliably test that requests pass through the proxy, and should also be useful for testing other HTTP related behavior. I also switched proxy configuration from using `http[s].proxyHost` to Hato's `:proxy` `java.net.ProxySelector`, which can be reset in tests (unlike `http.proxyHost`, which is resolved only once).

While doing this, I went overboard and added unit tests for every Hato based request in eca. I believe these tests add value beyond proxy authentication itself, but I'm happy to unwind or simplify anything if it makes the review easier.

Finally, I introduced a `--proxy` flag to the `integration-test` bb task. It spins up a `Tinyproxy` instance, sets the proxy environment variables, and verifies that all requests go through the proxy by checking for an additional header in the mock server (otherwise the request is rejected with 403). This flag is enabled in the GH workflow when testing the binary on Ubuntu.

In summary the changes are:

- Parse `http[s]_proxy` / `HTTP[S]_proxy` env vars in `proxy.clj`, producing a map with `:host`, `:port`, `:username`, and `:password`. Support both lower and uppercase names because they're case sensitive on MS-Windows.
- Introduce `client_http.clj`, which turns that map into Hato `:proxy` and `:authenticator` options, builds an HTTP client, and exposes it via a dynamic `*hato-http-client*` bound as the root client.
- Update `main.clj` to initialize `*hato-http-client*` with the correct proxy configuration.
- Ensure every Hato request uses `:http-client (client/merge-with-global-http-client http-client)`, combining the global proxy client with any per-request options (including custom `httpClient` providers).
- Add `client_test_helpers.clj` with a `with-client-proxied` macro to route requests through an embedded [LittleProxy](https://github.com/adamfisk/LittleProxy) instance for unit tests, returning controlled fake responses. Add LittleProxy as a `:test` dependency (and update `deps-lock.json`).
- Add unit tests using this macro for all functions that issue Hato requests, significantly increasing unit test coverage.
- Extend the `integration-test` bb task with the `--proxy` command line option, which sets `http_proxy` and verifies routing through [Tinyproxy](https://tinyproxy.github.io/). Update the GH workflow to `ubuntu-24.04` to support Tinyproxy.
- Add more integration-test options: `--dev` to run from source, `--list-ns` to list all test namespaces and `--ns` to restrict the namespaces executed.

I left out the changelog and documentation for now, pending feedback on the above.


scripts/make.clj
```diff
@@ -92,18 +93,59 @@
...
+(def integration-test-cli-spec
+  "Command-line options supported by the `integration-test` task."
+  {:spec
+   {:dev {:desc "Invoke the ECA server from the source files using the `clojure` CLI rather than the ECA binary."}
+    :help {:desc "Display help."
+           :alias :h}
+    :list-ns {:desc "List available test namespaces."
+              :alias :l}
+
+    :ns {:desc "Comma-separated list of test namespaces to run.."
```
ericdallo: cool!

ericdallo (commented):
Changes look good!
Let's discuss the issue first next time before the PR because it can cause re-work depending on the subject/planned changes, especially for big PRs like this one
LMK when you do the changelog changes and docs, thank you!

ikappaki:
> Changes look good! Let's discuss the issue first next time before the PR because it can cause re-work depending on the subject/planned changes, especially for big PRs like this one LMK when you do the changelog changes and docs, thank you!

I updated the release notes and added a proxy section to the config docs.

Sorry about the large patch size. It kind of blew up as I worked on it, especially around testing. I even had a version that let HTTP/HTTPS use different credentials, but it needed an extra `.java` file to bypass protected method overrides in the `Authenticator` class, so I scrapped it. I’ll try to give a heads up sooner next time if it starts growing like that.

Happy to support anything that pops up from the proxy work.

Thanks

PR #264, opened 2026-01-13
Fix gemini OpenAI chat tool calling 2

zikajk:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix Gemini (OpenAI compatible). #247
+
 ## 0.91.2
 
 - Fix `eca_shell_command` to include stderr output even when exit 0.


zikajk:
@ericdallo  Do you think the failing build can be related to my changes?

ericdallo:
no, I did a mistake in master that I'm taking a look, I let you know when fixed

src/eca/llm_providers/openai_chat.clj
```diff
@@ -25,13 +25,12 @@
...
   "Extract text content from various message content formats.
-   Handles: strings (legacy eca), nested arrays from chat.clj, and fallback."
+   Returns a string for text-only content, and an array only when images are present (multimodal)."
   [content supports-image?]
   (cond
     ;; Legacy/fallback: handles system messages, error strings, or unexpected simple text content
     (string? content)
-    [{:type "text"
-      :text (string/trim content)}]
+    (string/trim content)
```
ericdallo, line 33: I used maps to keep consistency in the data structure, but I'm ok if that fix problems, I just hope to not cause any other

zikajk: It is supported, but often not recommended so I decided to take the safe route here :-). 
And I suppose it will save some tokens as well.

zikajk:
@ericdallo The Gemini is still buggy though. It looks like its OPENAI api isn't perfect. 
E.g. it sends us `though>` with missing `<` which is fed back to the model and it gets very confused. 

https://github.com/charmbracelet/crush/issues/1698

image
image

zikajk:
@ericdallo Regarding the rare problem with broken gemini thinking tag. It is not something I will fix in this PR.

src/eca/llm_providers/openai_chat.clj
ericdallo: I don't think tokens are calculated based on the structure, but on the chars content, but I'm ok with that

ericdallo:
@zikajk does that means it's a bug only with gemini or other providers? is it possible to repro with other providers so I can try?

zikajk:
> @zikajk does that means it's a bug only with gemini or other providers? is it possible to repro with other providers so I can try?

Only with Gemini. I have an experimental branch where reasoning is purged from history (user can opt-out via provider config). 
And what I have found so far, it is quite possible that Gemini (and maybe other models as well?) doesn't even like to have reasoning returned in the same turn (nothing in official docs though...). 
I want to see if the problem is gone when i won't send ANY reasoning back. If yes, then I would send "turn reasoning" only for models doing delta-reasoning (Deepseek) - again with a opt-out via provider config.

ericdallo:
>  it is quite possible that Gemini (and maybe other models as well?) doesn't even like to have reasoning returned in the same turn (nothing in official docs though...).

I see, but for other models I'd say this is unlikely, as at least major models like openai and anthropic this is required, but let's test, but I'd say it should affect model quality

zikajk:
> > it is quite possible that Gemini (and maybe other models as well?) doesn't even like to have reasoning returned in the same turn (nothing in official docs though...).
> 
> I see, but for other models I'd say this is unlikely, as at least major models like openai and anthropic this is required, but let's test, but I'd say it should affect model quality

Can you point me to the documentation? I know they use different API but I am curious.

ericdallo:
> Can you point me to the documentation? I know they use different API but I am curious.

https://platform.openai.com/docs/api-reference/responses/create
>  Includes an encrypted version of reasoning tokens in reasoning item outputs. This enables reasoning items to be used in multi-turn conversations when using the Responses API statelessly (like when the store parameter is set to false, or when an organization is enrolled in the zero data retention program).

Anthropic has the same [here](https://platform.claude.com/docs/en/build-with-claude/extended-thinking), mentioning to pass back thinking blocks

zikajk:
@ericdallo 

> > Can you point me to the documentation? I know they use different API but I am curious.
> 
> https://platform.openai.com/docs/api-reference/responses/create
> 
> > Includes an encrypted version of reasoning tokens in reasoning item outputs. This enables reasoning items to be used in multi-turn conversations when using the Responses API statelessly (like when the store parameter is set to false, or when an organization is enrolled in the zero data retention program).
> 
> Anthropic has the same [here](https://platform.claude.com/docs/en/build-with-claude/extended-thinking), mentioning to pass back thinking blocks

The Gemini also requires to keep the encrypted version (thought_signatures). 
And the Anthropic behavior is the one I consider to be the best default. Only this part is something I wonder about (if other providers throws past thinking tokens away): 

> While you can omit thinking blocks from prior assistant role turns, we suggest always passing back all thinking blocks to the API for any multi-turn conversation. The API will:
> 
> Automatically filter the provided thinking blocks
> Use the relevant thinking blocks necessary to preserve the model's reasoning
> Only bill for the input tokens for the blocks shown to Claude

But yeah, I am keeping the mentioned branch experimental and I don't plan it to make it PR until I am absolutely sure that it helps. 
Not related to this branch though :-)

ericdallo (approved):
Is this done? can you add changelog entry please?

zikajk:
@ericdallo Done.

PR #275, opened 2026-01-20
Optimize large shell tool output

bobrowadam:

## Summary

- **Optimize large shell output**: When shell command output exceeds a configurable threshold (default: 1000 chars), write to a cache file (`~/.cache/eca/{workspace-hash}/shell-output/`) instead of returning inline. Reduces LLM context bloat for commands like `find`, `git log`, etc.
- **Auto-allow reading cache files**: Shell output cache files don't require additional read approval since the original command was already approved.

## Details

- New config option: `toolCall.shellCommand.outputFileThreshold` (default: 1000)
- Response includes file path, size, exit code, and last 20 lines for context
- Graceful fallback to inline output if file write fails
- Works for both success (exit 0) and error (non-zero exit) cases

### DIFF
--- a/test/eca/features/tools/shell_test.clj
+++ b/test/eca/features/tools/shell_test.clj
@@ -272,9 +272,11 @@
   (testing "Returns correct cache directory path"
     (let [workspaces [{:uri "file:///path/to/workspace"}]
           dir (#'f.tools.shell/shell-output-cache-dir workspaces)
-          dir-str (str dir)]
-      (is (string/includes? dir-str ".cache/eca"))
-      (is (string/includes? dir-str "shell-output")))))
+          dir-str (str dir)
+          ;; Normalize to forward slashes for cross-platform test
+          normalized-str (string/replace dir-str "\\" "/")]
+      (is (string/includes? normalized-str ".cache/eca"))
+      (is (string/includes? normalized-str "shell-output")))))
 
 (deftest eca-shell-output-cache-file-test
   (testing "Returns true for paths in the ECA shell output cache directory"


src/eca/features/tools/shell.clj
```diff
@@ -49,7 +53,62 @@
                         :continue true}
                  input (assoc :in input)))))
 
-(defn ^:private shell-command [arguments {:keys [db tool-call-id call-state-fn state-transition-fn]}]
+(defn ^:private workspaces-hash
```
ericdallo, line 56: Why are we duplicating this function if it exists in `eca.cache/workspaces-hash`?

src/eca/features/tools/shell.clj
```diff
@@ -49,7 +53,62 @@
...
+                   (sort))
+        joined (string/join ":" paths)
+        md (MessageDigest/getInstance "SHA-256")
+        digest (.digest (doto md (.update (.getBytes ^String joined "UTF-8"))))
+        encoder (-> (Base64/getUrlEncoder)
+                    (.withoutPadding))
+        key (.encodeToString encoder digest)]
+    (subs key 0 (min 8 (count key)))))
+
+(defn ^:private shell-output-cache-dir
```
ericdallo, line 71: I wonder if this functoin is better to live in eca.cache along with the `workspace-cache-file`

bobrowadam:
> @bobrowadam thank you for the help, but I wish we could discuss this better in a issue before a PR, I believe @zikajk has some concerns about this approach

@ericdallo  sure no problem. I'll be happy to be part of a discussion regarding optimizing token usage this way or another.

bobrowadam:
closing this PR for now

PR #276, opened 2026-01-21
Fix flaky basic-tool-calling-prompt-test

bobrowadam:
## Summary
Fix flaky `basic-tool-calling-prompt-test` that was causing intermittent CI failures.

The test was expecting a strict order of messages, but there's a race condition between progress messages ("Calling tool", "Generating") and tool state messages (toolCallRunning, toolCalled). Changed from strict ordering to using `m/in-any-order` which checks that all expected messages are present without requiring a specific order.

N/A - test-only change that doesn't affect end users.

### DIFF
--- a/test/eca/features/chat_test.clj
+++ b/test/eca/features/chat_test.clj
@@ -285,22 +285,25 @@
                                                                                          :type :text}]}}}
                                 {:role "assistant" :content [{:type :text :text "I can see: \n/foo/bar"}]}]}}
            (:chats (h/db))))
+      ;; Note: We use m/in-any-order because there's a race between progress messages
+      ;; ("Calling tool", "Generating") and tool state messages (toolCallRunning,
+      ;; toolCalled) - their relative order in the middle section is non-deterministic.
       (is (match?
            {:chat-content-received
-            [{:role :user :content {:type :text :text "List the files you are allowed to see\n"}}
-             {:role :system :content {:type :progress :state :running :text "Waiting model"}}
-             {:role :system :content {:type :progress :state :running :text "Generating"}}
-             {:role :assistant :content {:type :text :text "Ok,"}}
-             {:role :assistant :content {:type :text :text " working on it"}}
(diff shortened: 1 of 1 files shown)


test/eca/features/chat_test.clj
```diff
@@ -285,22 +285,24 @@
...
-             {:role :assistant :content {:type :toolCallPrepare :id "call-1" :name "list_allowed_directories" :arguments-text ""}}
-             {:role :assistant :content {:type :toolCallRun :id "call-1" :name "list_allowed_directories" :arguments {} :manual-approval false}}
-             {:role :assistant :content {:type :toolCallRunning :id "call-1" :name "list_allowed_directories" :arguments {}}}
-             {:role :system :content {:type :progress :state :running :text "Calling tool"}}
-             {:role :assistant :content {:type :toolCalled :id "call-1" :name "list_allowed_directories" :arguments {} :total-time-ms number? :outputs [{:text "Allowed directories: /foo/bar" :type :text}]}}
-             {:role :system :content {:type :progress :state :running :text "Generating"}}
-             {:role :assistant :content {:type :text :text "I can see: \n"}}
-             {:role :assistant :content {:type :text :text "/foo/bar"}}
-             {:role :system :content {:state :finished :type :progress}}]}
+            (m/embeds [{:role :user :content {:type :text :text "List the files you are allowed to see\n"}}
```
ericdallo: I believe m/embeds is not doing what you expect and can cause the test to falsely pass even if breaking, m/in-any-order is what you want, can you test it?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix flaky `basic-tool-calling-prompt-test` caused by race condition in message ordering.
```
ericdallo: we don't need changelog for tests changes that do not affect final user

bobrowadam: will remove that

test/eca/features/chat_test.clj
bobrowadam: will do

bobrowadam:
@ericdallo  Fixed PR comments

PR #295, opened 2026-02-09
Use models.dev as dynamic model source and harden model selection

### LINKED ISSUE #292: Add model discovery through Models.dev

When a provider API URL (or name) matches Models.dev and provider `:fetchModels != false` then all available models will be fetched automatically.

zikajk:
- switch dynamic model discovery to models.dev with fetchModels opt-out
- merge static model config with discovered defaults and preserve modelName aliases
- normalize/join provider API URLs to avoid malformed endpoints
- make default model selection deterministic and fail fast when no model is available
- update tests and docs for the new model-loading behavior

Closes #292

- [x] This is not an AI slop.

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -3,35 +3,50 @@
   "$id": "https://eca.dev/config.json",
   "title": "ECA Configuration",
   "description": "Configuration schema for ECA (Editor Code Assistant). See https://eca.dev/configuration for full documentation.",
+  "markdownDescription": "Configuration schema for ECA (Editor Code Assistant). See https://eca.dev/configuration for full documentation.",
   "type": "object",
   "additionalProperties": false,
   "allowComments": true,
   "properties": {
     "$schema": {
       "type": "string",
-      "description": "JSON Schema reference for editor validation support."
+      "description": "JSON Schema reference for editor validation support.",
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -676,7 +676,135 @@ To configure, add your OTLP collector config via `:otlp` map following [otlp aut
 }
 ```
 
-## Default config values
+## All configs
```
ericdallo: we removed this section in a previous commit in master, now we have a dedicated config.json schema like opencode

zikajk: Thanks, I did a mistake when rebasing. Fixed.

src/eca/shared.clj
```diff
@@ -28,6 +28,24 @@
   "The system's line separator."
   (System/lineSeparator))
 
+(defn normalize-api-url [api-url]
+  (some-> api-url
+          string/trim
+          (string/replace #"/+$" "")))
+
+(defn join-api-url
```
ericdallo, line 36: this one is good, there are people that had issues with adding `/` to the end of the api url and eca didn't work

ericdallo:
Fixed there!

zikajk:
@ericdallo Thank you :-)

PR #305, opened 2026-02-14
Fix NullPointerException when model is nil in chat and completion

djgoku:
**Note: I am not a closure developer nor did I write this claude did, but this seemed like a reasonable change.**

I wanted to try out eca on emacs, but from a clean installation using mise eca (which doesn’t seem to matter), I encountered [eca-chat-error.txt](https://github.com/user-attachments/files/25317683/eca-chat-error.txt) and [eca-chat-error-log.txt](https://github.com/user-attachments/files/25317682/eca-chat-error-log.txt).

After making these changes, I no longer encounter this issue with a locally built native-cli binary: [eca-chat.txt](https://github.com/user-attachments/files/25317681/eca-chat.txt)

- Extract parse-model utility to llm-util with nil safety via some->, replacing inline string/split calls across chat.clj, completion.clj, rewrite.clj, and models.clj.
- Add early return in completion when no model is configured. Add tests for nil model scenarios.

- [] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix NullPointerException when model is nil in chat and completion
+
 ## 0.101.0
 
 - Foreground subagents support! #160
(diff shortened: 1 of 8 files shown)


ericdallo:
Thank you for the help, but I just fixed that on latest release a few mins ago 😅 LMK if works for you!

djgoku:
Yup just saw this lol https://github.com/editor-code-assistant/eca/commit/ab48040439fbf0029996a408180411351c1a6d64

djgoku:
@ericdallo it does work thanks.

PR #311, opened 2026-02-16
Improve plan/explorer delegation workflow and read-only allowlist

zikajk:
Following up on our discussion regarding the Explorer subagent.

The current instructions have an issue where they conflict with the General agent’s instruction—specifically the part about 'researching complex questions.' The problem isn't just the conflict itself, but the fact that if the Explorer is preferred for deep research and then tries to use a skill intended to write findings to a file, it crashes. Because of this, I would still let the General subagent handle the vast majority of tasks.

The Explorer subagent makes sense in Plan mode, for example. Previously, the EXPLORE phase could eat up half the context window; that is now resolved (though there may still be rare cases where using a subagent isn't necessarily worth it or where model ignores instructions, but I’d say those will be minimal).

Without the Explorer, my plan consumed around 100k tokens. With this new delegated aggressive approach and the changes to Plan mode, I managed to get it down to 27k tokens. Personally, I see the Explorer’s greatest strength in how it handles browsing files, diffs, and similar tasks.

I’m curious to hear your thoughts on this. Give it a try, please:-) 

- [x] This is not an AI slop.

### DIFF
--- a/test/eca/features/tools_test.clj
+++ b/test/eca/features/tools_test.clj
@@ -242,11 +242,11 @@
         ;; Test git operations
         (is (= :deny (f.tools/approval all-tools shell-tool {"command" "git add ."} {} config "plan")))
         ;; Test safe commands that should NOT be denied
-        (is (= :ask (f.tools/approval all-tools shell-tool {"command" "ls -la"} {} config "plan")))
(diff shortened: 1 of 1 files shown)


ericdallo:
Tests are failling because of toolcall approval changes, but I will fix on master

zikajk:
@ericdallo Thank you. I am happy that you like it. 
It worked nice in my tests too.

PR #320, opened 2026-02-19
Fix disabledTools ignoring mcp server name prefix

zikajk:
- [x] This is not an AI slop.


ericdallo:
I'm fixing the unit test at master

PR #356, opened 2026-03-11
update Github copilot instructions

CsBigDataHub:
- [X] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## Unreleased
 
-- Update Github copilot instructions on how to configure for different copilot plans.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Update Github copilot instructions on how to configure for different copilot plans.
```
ericdallo: We don' t need this as it won't affect next release
```suggestion
```

ericdallo (approved):
Interesting, for now I think we can have this doc but we may want to improve /login to ask that for user.
WDYT @zikajk ?

PR #406, opened 2026-04-12
Refresh auth before each LLM API call, not just at prompt start

zikajk:
Provider auth was captured once at prompt start and reused across retries and within-stream tool rounds. If a token expired during a long-running tool call (e.g. spawn_agent), all subsequent requests used the stale key.

Two-layer fix:
- Layer 1 (llm_api): refresh-provider-auth-fn called before each prompt! invocation, covering initial calls and retries
- Layer 2 (tool_calls): with-fresh-auth helper ensures all continuation branches (including rejections) propagate refreshed credentials back to the provider's streaming loop

Also restores missing messages-after-last-compact-marker in the no-continue-fn fallback path.

- [ ] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Refresh auth token before each LLM API call, preventing stale tokens during long-running tool calls.
+
 ## 0.124.5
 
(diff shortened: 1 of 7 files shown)


ericdallo (approved):
Looks good!

PR #412, opened 2026-04-14
Add chat `/model` command

SGospodinov:
Let users switch current chat model directly from slash commands, while keeping model selection synced with client state and preserving chat-level variant behavior across prompts and forks.

🤖 Generated with [eca](https://eca.dev/)

Co-Authored-By: eca-agent <git@eca.dev>

- [ ] This is not an AI slop.

I ran the test did manual testing for the compiled server to validate the results and it seems to be working as intended, but the code is generated using the ECA agent. Cannot guarantee that results are not slop. I'm new to the ECA project and Clojure and I'm using the PR to gauge the interest in the feature and to learn more about the way ECA work. Will understand if the PR gets closed because the human input was low and the code doesn't match the necessary standards.


src/eca/features/chat.clj
```diff
@@ -1063,6 +1063,11 @@
                       new-id))
         selected-agent (config/validate-agent-name raw-agent config)
         agent-config (get-in config [:agent selected-agent])
+        chat-state (get-in @db* [:chats chat-id])
```
ericdallo: why we need the changes in this file? it doesn't look it's related to the feature of this PR

SGospodinov: My bad, that was leftover from an earlier version allowing you to also select model variant with the a single call. It's not necessary, will remove it.

ericdallo (approved):
Thank you!

PR #429, opened 2026-04-26
Improve rules with filters, conditions, and scoped loading

zikajk:
- Add YAML frontmatter support for filtering rules by agent, model, paths, and enforcement mode.
- Split rules into static prompt-injected rules and path-scoped rules that can be loaded on demand with `fetch_rule`.
- Add condition-variable rendering, `/rules` inspection, docs, and tests for the expanded rules behavior.

- [x] This is not an AI slop.


ericdallo (approved):
💯

PR #436, opened 2026-04-28
Support regex patterns for shell_command in markdown agent tools

zikajk:
Add `tool(regex)` syntax to markdown agent allow/ask/deny lists to restrict approvals by argument pattern. Currently limited to `eca__shell_command` (matched against its `command` arg); on other tools the pattern is ignored with a warning. Repeated entries for the same tool are merged into one argsMatchers list. For broader argsMatchers use (other tools, multiple args), use JSON `toolCall` config

- [x] This is not an AI slop.


ericdallo (approved):
Looks good

PR #442, opened 2026-05-02
Detect command and skill args from $ARGN/$N/$ARGS placeholders

zikajk:
Command and skill files previously exposed hardcoded :arguments [] in the command listing regardless of their content. Now they scan for $ARG1, $1, $ARGS, and $ARGUMENTS placeholders and produce matching :arguments metadata so clients can prompt users correctly.

Native commands that require arguments (skill-create, plugin-install, plugin-uninstall) are now marked :required true.

Deduplicate prefixed-command-name and prefixed-skill-name into shared/prefixed-name. Fix protocol docs to mark required as optional.

- [x] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Custom commands and skills now expose `:arguments` metadata inferred from their content. Previously they always reported empty arguments.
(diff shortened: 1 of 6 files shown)


ericdallo (approved):
nice

PR #450, opened 2026-05-05
Resolve bare native ECA tool names

zikajk:
Some LLM tool calls occasionally omit the `eca__` prefix for native tools, such as `write_file` instead of `eca__write_file`.
This caused native tool dispatch to fail and could lead to repeated retries before the model recovered.`

Resolve bare native ECA tool names to their canonical form, keep MCP tool names strict, and log auto-resolved native tool calls for debugging.

- [x] This is not an AI slop.


zikajk:
@ericdallo This might be a bit controversial :sweat_smile: , but some models still try to call tools without a prefix first. For a while, I thought about extending this to MCP if the tool has a unique name and there's no risk of collision, but I ultimately decided to stick with native tools only. I haven't noticed any regressions during several hours of testing, so I believe this is a safe change (and also consistent with what we allow in the config).

ericdallo:
@zikajk hum, what models you saw that?
Also CI tests are failing

zikajk:
@ericdallo Deepseek 4 and local QWEN.

ericdallo:
Ok, let's go ahead since this seems safe enough and see how it goes

zikajk:
@ericdallo Tests are fixed, the error isn't related to changes here which is interesting as other pipelines were green. 
Maybe it would be a good idea to use m/embed and make these tests less strict so they don't fail when the description changes or something like that. But let's leave that for another time :-)

ericdallo (approved):

PR #451, opened 2026-05-07
Support Claude console subscription auth for model listing

bobrowadam:
- Dispatch `x-api-key` vs `Authorization: Bearer` when fetching the provider `/models` endpoint based on the stored auth `:type`, so Claude console subscription (token) and OAuth flows both work.
- Thread `auth-type` from `provider-api-key` through `fetch-provider-native-models` instead of hardcoding `:auth/oauth`.
- Add tests for both Anthropic token (`x-api-key`) and OAuth (`Bearer`) paths.

- [x] This is not an AI slop.


ericdallo:
@bobrowadam what this fixes? what you see? can you check tests as well?

ericdallo:
I get the issue now, but can you fix the tests before I test?

bobrowadam:
@ericdallo sure will do

ericdallo (approved):
Thank you!

PR #452, opened 2026-05-08
Fix OpenAI OAuth web search and image generation

itkonen:
The OpenAI built-in web-search and image-generation didn't work when authentication with OAuth, because the authentication method flag was used to drop this feature.

New issues arised as the newer models required the `web_search` tool instead of the legacy `web_search_preview` tool. Also, the server tried to send web-search artefacts back to OpenAi when continuing the conversation, which caused an API error.

This PR should fix these issues. I got openai/gpt-5.5 and openai/gpt-5.4 models with OAuth to search the web and generate images. openai/gpt-5.3-codex doesn't see the tools as per design. I couldn't verify if some older models using API key authentication might still rely on the `web_search_preview` tool.

---- 

## AI Summary 
- Preserve OpenAI built-in web search and image generation tools when using OAuth authentication, relying on model capabilities instead of auth type to decide tool availability.
- Use the current OpenAI Responses API `web_search` tool type instead of the legacy `web_search_preview` type.
- Skip OpenAI server web-search history artifacts during replay so follow-up turns after web search do not crash.

## Verification
- `clojure -M:test --focus eca.llm-providers.openai-test`
- Manual test with a locally built ECA server confirmed OpenAI OAuth web search works across follow-up turns.
- `clj-kondo --lint src/eca/llm_providers/openai.clj test/eca/llm_providers/openai_test.clj` could not be run locally because `clj-kondo` is not installed in this environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix OpenAI OAuth requests dropping built-in web search and image generation tools, and avoid replaying web search history artifacts on follow-up turns.
+
 ## 0.133.1
 
 - Support client-generated chat ids: clients may now create the `chatId` themselves and send it on the first `chat/prompt`. eca-emacs#231.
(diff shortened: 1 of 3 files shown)


ericdallo (approved):
Thank you!

PR #459, opened 2026-05-11
Replace stderr-print logger with Logback/SLF4J and MDC context

zikajk:
- Structured logging via SLF4J/Logback: timestamps, log levels, chat-id MDC context
- Third-party noise suppression: root logger at WARN, eca at INFO
- MDC context propagation across future* threads
- Route MCP traffic through Logback (replaces raw eprintln from plumcp default)
- Fix --verbose: JSON-RPC tracing now passes at INFO independently of --log-level
- Fix --verbose CLI description (was misleading about stdout vs stderr)

- [x] This is not an AI slop.


ericdallo (commented):
- I think `prompt-steer-remove` is missing the `with-chat-context`, probably related to you implementing before I added that and you missed in a merge
besides that, looks pretty solid

zikajk:
@ericdallo Thanks! I also improved AGENTS.md so the clanker is aware of it.

ericdallo (approved):

PR #473, opened 2026-05-28
Add read-chat CLI command and centralize message sanitization

zikajk:
New `eca read-chat` subcommand streams raw chat records from the DB cache as JSONL — listing chats or filtering messages by time, role, and chat-id — without requiring a running server.

To support read-chat's time filtering, every message now receives a :created-at timestamp at insertion (backfilled for compact markers and flags). The new eca.message-sanitize namespace centralizes stripping of internal metadata (:created-at, :content-id) from messages before they reach LLM providers, replacing per-provider (dissoc :content-id) calls with a single sanitize-outbound-messages entry point called from llm-api and each provider's tool-call loop.

- [x] This is not an AI slop.

### DIFF
--- a/src/eca/read_chat.clj
+++ b/src/eca/read_chat.clj
@@ -1,0 +1,190 @@
+(ns eca.read-chat
+  (:require
+   [clojure.java.io :as io]
+
+   [babashka.cli :as cli]
+   [cheshire.core :as cheshire]
+
+   [eca.cache :as cache]
+   [eca.db :as db]
+   [eca.shared :as shared])
+  (:import
+   [java.time Instant LocalDate ZoneOffset]))
+
+(set! *warn-on-reflection* true)
+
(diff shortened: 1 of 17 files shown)


src/eca/read_chat.clj
```diff
@@ -0,0 +1,190 @@
+(ns eca.read-chat
```
ericdallo, line 1: should this be eca.features.read-chat?

zikajk: I was thinking it would be nice to separate the top-level CLI stuff from the internal features, which is why I left it here. What do you think?

ericdallo: yes, I think that's fine, almost like public api right? I wonder how this wil scale tho, but refactoable in the future

ericdallo (approved):
Thank you!

PR #479, opened 2026-05-30
Fix Anthropic Streaming Responses

bobrowadam:
## Problem

Anthropic streaming responses arrive all at once at the end instead of
token-by-token. Anthropic gzip-compresses the SSE body and `hato` decompresses
it, but `GZIPInputStream` only emits bytes once a full deflate block arrives,
so the stream is buffered until the response is nearly complete.

## Fix

Send `accept-encoding: identity` and set hato's
`:decompress-body false` so the event-stream stays uncompressed and can be read
line-by-line.

## Tradeoff
SSE text is no longer gzipped, but for interactive use it's negligible and time-to-first-token improves.
Server-side chunk granularity is unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


ericdallo (approved):
Interesting, thanks for that!

PR #482, opened 2026-06-02
Fix Anthropic 400 on empty thinking blocks in multi-turn chats

### LINKED ISSUE #481: Anthropic 400: thinking field is null when thinking block had no text deltas
When using anthropic models via anthropic, I started getting these errors today after first "thinking". I switched to using sonnet-4-6 via github-copilot and I just asked it to investigate, but it also submitted this on my behalf before I had time to test further. I have now tested the fix locally and it seems to work. I'll keep using it and report if the problem repeats.

## Bug

When a conversation contains a stored `reason` message from an Anthropic thinking block that accumulated zero `thinking_delta` events (i.e. the model returned an empty thinking block), ECA sends `null` for the `thinking` field on the next request. Anthropic's API rejects this with:

```
Anthropic response status: 400 body: {"type":"error","error":{"type":"invalid_request_error","message":"messages.3.content.0.thinking.thinking: Input should be a valid string"},"request_id":"req_011Cbe6NenQJS91qdUN3DS3L"}
```

## Root cause

In `src/eca/llm_providers/anthropic.clj`, when serializing a stored `"reason"` message back into an outbound request:

```clojure
{:type "thinking"
 :signature (:external-id content)
 :thinking (:text content)}   ; <-- nil when no thinking_delta was received
```

`(:text reasoning)` starts as `nil` in the `reasonings*` atom and is only set via:

```clojure
:thinking (do (swap! reasonings* update-in [id :text] str text) ...)
```

If no `:thinking` delta arrives before `:finished` (empty thinking block), `(:text reasoning)` remains `nil`.

## Fix

Change line ~242 from:

```clojure
:thinking (:text content)
```

to:

```clojure
:thinking (or (:text content) "")
```

I cloned the repo, built locally, copied patched eca server, restarted server, resumed a chat that previously today everytime ended in the error, and now it works. Also new chats go beoynd two "thinking"s.

## Environment

- ECA version: 0.136.4
- Provider: anthropic (direct)
- Model producing the error: claude-* with extended thinking enabled

avehtari:
- [ ] This is not an AI slop.

The fix and regerssion test were generated by Claude. I don't think this is slop, but you can decide. I have read the new code and I understand the fix. I have tested that all unit tests pass, new regression test, and I have been running the patched server succesfully.

## Summary

- Direct Anthropic chats fail on the 2nd+ turn with `400 ... messages.N.content.0.thinking.thinking: Input should be a valid string` when a prior turn produced a thinking block with no content.
- Root cause: a `content_block_start` of type `thinking` with no `thinking_delta` before `content_block_stop` leaves `(:text content)` as `nil`. When that `reason` message is replayed in `normalize-messages`, `:thinking` is serialized as `nil`, which Anthropic rejects.
- Fix: default to `""` — `:thinking (or (:text content) "")` in the `reason` branch of `normalize-messages`.

## Test

Adds `reason-empty-thinking-test` covering nil `:text`, missing `:text` key, present `:text` (preserved), and redacted thinking (unaffected). Verified it fails on the unpatched code (`nil` vs `""`) and passes with the fix. Full suite: 593 tests, 3225 assertions, 0 failures.

Closes #481

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Remote REST API: `GET /api/v1/chats/:id` now exposes the chat's effective `model`, `variant` and `agent` (per-chat override, falling back to the resolved session default).
+- Anthropic: serialize empty thinking blocks as `""` instead of nil, fixing a 400 on multi-turn chats with extended thinking. (#481)
 
 ## 0.136.4
 
--- a/src/eca/llm_providers/anthropic.clj
+++ b/src/eca/llm_providers/anthropic.clj
@@ -239,7 +239,7 @@
                               :data (:data content)}
                              {:type "thinking"
                               :signature (:external-id content)
-                              :thinking (:text content)})]})
+                              :thinking (or (:text content) "")})]})
 
               "server_tool_use"
               (when-not foreign-api?
--- a/test/eca/llm_providers/anthropic_test.clj
+++ b/test/eca/llm_providers/anthropic_test.clj
@@ -240,6 +240,41 @@
                        {:type :text :text "Here are the results."}]}]
            result)))))
 
+(deftest reason-empty-thinking-test
+  (testing "reason message with nil :text serializes :thinking to \"\" (not nil)"
+    ;; Regression: Anthropic rejects requests where a replayed thinking block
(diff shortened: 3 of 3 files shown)


ericdallo (approved):
Thank you!

PR #492, opened 2026-06-05
Add OpenAI Responses API for Copilot's gpt-5.5 and gpt-5.4-mini

zikajk:
- [x] This is not an AI slop.


zikajk:
@ericdallo 
Please. With the new changes in Copilot, gpt-5.4-mini will probably be a pretty popular model :sweat_smile:

PR #493, opened 2026-06-06
Surface waiting ask_user tool calls in remote pendingToolCalls

rschmukler:
A web/SSE client that reconnects rebuilds state via GET /api/v1/chats/:id, but a chat blocked on ask_user was invisible there: the tool call sits in :executing (not :waiting-approval) while its handler blocks on the answer, so pendingToolCalls never listed it and the client couldn't re-render the question.

ask_user now mints its requestId up front, stores it in tool-call state, and passes it to the messenger so the SSE event and the persisted id match. pendingToolCalls includes :executing ask_user calls and carries requestId so a reconnecting client can answer via POST /api/v1/answer.

- [x] This is not an AI slop.


ericdallo (approved):
thank you!

PR #494, opened 2026-06-06
Deliver ask_user to both editor and SSE clients concurrently

rschmukler:
This PR builds on #493 to change the behavior of eca such that it will now broadcast to both the editor and SSE clients. This is useful so that if you happen to have an eca client open (browser UI, iOS client, etc) you don't make it so your editor is incapable of responding to `ask_user`.

Clients should be updated to handle closing the ask_user dialog when they see a `toolCalled` or `toolCallRejected` notification with the same `toolCallId` (and assumed it was handled by a different client). This is already consistent with what clients are expected to do with other tool calls.

- [x] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


ericdallo (approved):
Nice catch ,thanks

PR #496, opened 2026-06-07
Add authScope to namespace MCP OAuth tokens per project

rschmukler:
MCP OAuth tokens were stored in the global cache keyed only by server name, so two projects using a same-named server (e.g. a plugin-provided `linear`) shared one `[:mcp-auth <name>]` slot. Authenticating a different account in one project clobbered the other's token.

Introduce an optional `authScope` field on HTTP MCP server config and derive the storage key from it via a single `mcp-auth-key` helper that all reads/writes/deletes route through:
- `global` (default/unset): bare server name, one token shared everywhere (back-compat, no migration of existing tokens).
- `workspace`: scoped to the workspace folder set (reuses `cache/workspaces-hash`).
- any other value: a named bucket shared by projects using that value.

The server name always prefixes composite keys so different servers never share a token, and tool names (`<server>__<tool>`) are unaffected. Storage stays in the single global cache; only the key is namespaced.

Document the field in config.json, protocol.md and tools.md, including that it supports the full dynamic-string interpolation like other config values.

- [x] This is not an AI slop.


ericdallo (approved):
Not sure that's too much but indeed improve auth for mcps, I'm ok with that for now, thanks !

PR #498, opened 2026-06-12
Add opt-in message pagination to remote chat HTTP API

rschmukler:
GET /api/v1/chats/:id accepts limit/before/after query params with opaque cursors and an after=lastCompaction sentinel, returning a messagesMeta object with before/after/compaction cursors.

Add opt-in pagination via limit/before/after query params. When none are
present the response is unchanged, so existing clients keep working; the
new behavior is purely additive (a sibling messagesMeta object alongside
the existing messages array), which the object-shaped response already
accommodates without a breaking change.

Cursors are opaque base64url tokens encoding `<index>.<checksum>`, where
the checksum fingerprints the message. Resolution trusts the encoded
index when its checksum still matches, otherwise rescans by checksum, so
cursors survive mid-history shifts (e.g. flag insert/remove) and only a
genuinely removed message (rollback/clear) yields 409 cursor_expired,
prompting the client to refetch. Token pruning rewrites content in place
and never shifts positions, so the active window stays stable.

Pages are newest-anchored so opening a chat shows the latest context,
with before/after cursors computed against full history so paging
transparently crosses the compaction boundary. The after=lastCompaction
sentinel returns the active context since the last compaction in a single
call, fulfilling the original "messages since last compaction" goal
without a dedicated flag; before=lastCompaction returns the summarized
history. Cursors are nil at the ends (nil-punning) instead of separate
hasMore/hasBefore fields.

## Request examples

```http
# Newest 50 messages (with pagination metadata)
GET /api/v1/chats/abc123?limit=50

# Active context since the last compaction
GET /api/v1/chats/abc123?after=lastCompaction&limit=50

# Older page (scroll up) — use beforeCursor from a previous response
GET /api/v1/chats/abc123?before=eyJpZHgiOjEyfQ&limit=50

# Newer page (scroll down) — use afterCursor from a previous response
GET /api/v1/chats/abc123?after=eyJpZHgiOjYyfQ&limit=50

# The summarized-away history (before the last compaction)
GET /api/v1/chats/abc123?before=lastCompaction
```

## Response shape

```jsonc
{
  "id": "abc123",
  "title": "My chat",
  "status": "idle",
  "messages": [ /* ...the page... */ ],
  "messagesMeta": {
    "total": 412,                  // total messages in full history
    "returned": 50,                // messages in this page
    "beforeCursor": "eyJpZHgiOjEyfQ", // pass as `before` to load older (null at start)
    "afterCursor": null,           // pass as `after` to load newer (null at tail)
    "compactionCursor": "eyJpZHgiOjUwfQ" // boundary of last compaction (null if never compacted)
  }
  // ...model, agent, variant, pendingToolCalls, etc.
}
```

> With no `limit`/`before`/`after` params, the response omits `messagesMeta` and returns the full `messages` array (unchanged).

A stale cursor returns:

```jsonc
// 409 Conflict
{ "error": { "code": "cursor_expired", "message": "..." } }
```

- [x] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Remote HTTP `GET /api/v1/chats/:id`: opt-in message pagination via `limit`/`before`/`after` query params (opaque cursors, `after=lastCompaction` sentinel) returning `messagesMeta` with before/after/compaction cursors.
+
 ## 0.140.1
 
 - MCP: recover remote servers with stale connections (e.g. after suspend): tool-call timeouts now probe and re-initialize; add `mcpKeepAliveSeconds` pings; 404 triggers re-init.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Remote HTTP `GET /api/v1/chats/:id`: opt-in message pagination via `limit`/`before`/`after` query params (opaque cursors, `after=lastCompaction` sentinel) returning `messagesMeta` with before/after/compaction cursors.
+- Add message pagination over a shared cursor core: new JSON-RPC `chat/history` method and optional `limit`/`before`/`after` window params on `chat/open`, plus opt-in pagination on remote HTTP `GET /api/v1/chats/:id`. Opaque cursors with a `lastCompaction` sentinel; meta exposes before/after/compaction cursors.
 
 ## 0.140.1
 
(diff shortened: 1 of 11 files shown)


docs/config/remote.md
```diff
@@ -153,3 +153,34 @@ The web frontend provides a connect form where you enter the host and password (
 **Auto-discovery** — When you enter a host and click "Discover & Connect", the web UI scans ports `7777`–`7796` in parallel and finds all running ECA instances automatically.
 
 **Multi-session** — You can connect to multiple ECA instances simultaneously. Each connection appears as a tab in the top bar.
+
+## HTTP API: message pagination
```
ericdallo: I don't think we need to add any of this, this doc is for ECA users, not a explanation how ECA clients remote connection work, more to teach how users can use ECA remote feature.

ericdallo (approved):
looks good, also I checked with eca-emacs if it will fit good and looks good too.
I'll make the changes in editor clients starting with emacs

PR #499, opened 2026-06-13
Fix OpenAI OAuth context window detection

thanosapollo:
OpenAI OAuth uses the ChatGPT Codex backend, whose context
windows differ from the direct OpenAI API catalog.

ECA was resolving OAuth model limits like direct API-key auth, so
`openai/gpt-5.5` reported the direct API's 1.05M context window
instead of the Codex OAuth 272k limit.

This resolves OAuth limits from Codex `/models` first, with known
Codex OAuth caps as fallback. API-key OpenAI behavior is unchanged.

- [X] This is not an AI slop.


ericdallo:
Thanks for digging into this! We went with a different, provider-agnostic approach: instead of Codex-specific code in models.clj, there's now a generic provider-models-override multimethod in llm-util that the openai provider implements for OAuth (fetches Codex /models with known caps as fallback, returning standard limit overrides). Same fix for the gpt-5.5 window, but models.clj stays provider-agnostic. Landed on master in 945975849, so closing this one.

Issue #505, opened 2026-06-17
Subagents trigger postRequest hook after finishing


itkonen:
To upvote this issue, give it a thumbs up. See [this list](https://github.com/editor-code-assistant/eca/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc) for the most upvoted issues.

The [notify-hooks-plugin](https://plugins.eca.dev/plugins/notify-hooks/) is triggered afted a subagent finishes work, a regression since a 140.0, I assume. 

**To Reproduce**

1. Install the notify-hooks plugin
2. Prompt: "Spawn a subagent to say hello."
3. Hear the notification twice, once for the subagent, then again for the main agent.

**Expected behavior**

A notification when human interaction is required.

ericdallo:
Thank you, fixed!

PR #508, opened 2026-06-19
Keep postRequest primary-chat only

itkonen:
Issue #505 still remains after the previous fix. The remaining problem is that the `postRequest` hook is still fired by subagents, which can still cause duplicate or misleading post-request notifications.

This PR limits `postRequest` exclusively to main/primary-agent chats, and makes `subagentPostRequest` the exclusive post-request hook for subagent chats.

## AI summary

This change separates post-request hook dispatch by chat type:

- `postRequest` now runs only after primary-agent prompts.
- `subagentPostRequest` now runs only after subagent prompts.
- subagent hook payloads continue to include `parent_chat_id`.
- `continue: false`, `followUp`, `systemMessage`, and exit-2 follow-up behavior continue to work for the selected post-request hook type.

The lifecycle tests, hook documentation, and changelog have been updated to match the new contract.

Refs #505.

## Testing

- `clojure -M:test --focus eca.features.hooks-test`
- `LC_ALL=C bb test`
- `clj-kondo --lint src/eca/features/chat/lifecycle.clj test/eca/features/hooks_test.clj`

## Checklist

- [x] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Bugfix: `postRequest` hook no longer fires for subagent chats; subagents use `subagentPostRequest` only, preventing duplicate hook notifications. (#505)
 - Bugfix: clearer rejected tool-call result wording so models no longer assume a rejected edit was applied; it now states the call did not run and changed nothing. (#507)
 - Bugfix: `ask_user` normalizes the `options` argument so a malformed value (e.g. a stringified options an LLM sometimes emits) no longer reaches clients as broken choices. Accepts string/object arrays, recovers a JSON-encoded string, and drops unusable input.
 
--- a/docs/config/hooks.md
+++ b/docs/config/hooks.md
@@ -267,7 +267,7 @@ Fires before a prompt is sent to the LLM. Use for prompt validation, rewriting, 
 
 ### `postRequest`
 
-Fires after a primary-agent prompt finishes. Also runs for subagents. Primary use: validate the response or trigger a follow-up turn.
+Fires after a primary-agent prompt finishes. It does not run for subagent chats; use [`subagentPostRequest`](#subagentpostrequest) for subagent completions. Primary use: validate the response or trigger a follow-up turn.
 
 !!! note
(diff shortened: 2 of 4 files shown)


itkonen:
**CI failure is pre-existing and unrelated to this PR**

The `background-job-lifecycle` integration test failure is caused by a race condition in `src/eca/features/background_tasks.clj` — nothing touched by this PR.

**Root cause:** `kill-job!` calls `kill-job-impl` (sends SIGTERM) and *then* atomically sets `:status :killed`. The `monitor-process-exit!` daemon thread can race ahead: it dereferences the now-exited process, gets exit code 143, and calls `update-job-on-exit` before the kill status has been committed — overwriting it with `:failed` instead.

**Fix:** swap the order in `kill-job!` — set `:status :killed` in the registry *before* calling `kill-job-impl`, so the monitor thread always sees `:killed` and the guard in `update-job-on-exit` correctly skips the overwrite:

```clojure
(defn kill-job! [job-id]
  (if-let [job (get-job job-id)]
    (if (= :running (:status job))
      (do
        ;; Commit :killed status before destroying the process so that
        ;; monitor-process-exit! never overwrites it with :failed (exit 143).
        (swap! registry* (fn [reg]
                           (-> reg
                               (assoc-in [:jobs job-id :status] :killed)
                               (assoc-in [:jobs job-id :ended-at] (Instant/now)))))
        (kill-job-impl job)
        (logger/info logger-tag "Killed background job" {:id job-id})
        true)
      ...)))
```

The existing unit test `"does not overwrite :killed status"` in `background_tasks_test.clj` already covers this scenario.

ericdallo:
c/c @zikajk

zikajk:
@itkonen @ericdallo 
I said it was a feature, but it isn't. I got confused by some plugins, but this is actually a good change. Thanks :-).

itkonen:
Great this resolved! I already started orchestrating a fix around your earlier idea around `parent_chat_id`, which would have also worked fine. :) But I guess this first solution makes it simpler to separate main and sub-agent events if there are no compatibility issues.

zikajk:
@itkonen I was wrong, because Claude Code allows you to specify hooks within subagents frontmatter, and when you define a Stop hook there, it automatically changes to a subagentStop hook. But in all other cases (hooks in the global / project config), the Stop and subagentStop hooks are separate :-)

PR #509, opened 2026-06-19
Rework system-prompt cache identity and /prompt-show output`

zikajk:
The static prompt cache was keyed only on agent+model, so adding or removing file contexts silently reused stale instructions. Introduce a SHA-256 signature over all inputs that affect static prompt rendering (contexts, rules, skills, tools, environment, chat-prompt template); repo-map content is represented by presence only to avoid running `git ls-files` on every cache hit.

Consolidate SHA-256 helpers into eca.digest and reuse in cache.clj and oauth.clj. Quote-safe attribute rendering replaces ad-hoc string formatting in context/rule/skill XML tags.

/prompt-show is now read-only: it no longer consumes the editor-state cursor, and its output is reorganized into clearly delimited sections (Instructions / Chat) matching what the model receives. Fix a double blank line before ## MCP Resources caused by a trailing separator leaking when the Context section is absent.

- [x] This is not an AI slop.


ericdallo (approved):
Looks good, thank you!

PR #511, opened 2026-06-20
Include parent_chat_id in subagent postRequest hooks

itkonen:
## Summary

- Include `parent_chat_id` in subagent `postRequest` hook payloads.
- Document the subagent-specific field and cover it in hook lifecycle tests.
- Keeps `postRequest` valid for both primary and subagent chats while allowing hooks to distinguish them.

Follow-up to #505.

- [x] This is not an AI slop.

🤖 Generated with [ECA](https://eca.dev) (openai/gpt-5.5 - high)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Bugfix: subagent `postRequest` hooks now include `parent_chat_id`, letting hooks distinguish primary and subagent completions. (#505)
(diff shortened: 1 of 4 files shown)


itkonen:
Closing this because #508 was merged after the follow-up discussion clarified the intended behavior: `postRequest` should be primary-chat only, and subagents should use `subagentPostRequest`.

This PR was based on the earlier interpretation that `postRequest` should remain shared and expose `parent_chat_id`, so it now conflicts with the accepted direction.

PR #513, opened 2026-06-23
Fix MCP OAuth servers hanging on tool calls after session expiry

rschmukler:
When an MCP server's OAuth session expired or was revoked server-side (e.g. Linear), tool calls would hang for the full 120s timeout and never recover, instead of detecting the broken session and re-authenticating.

Root cause was in the Streamable HTTP transport: when the server returned a non-200 (401/403/404/5xx) whose body was not already a valid JSON-RPC message, plumcp synthesized a JSON-RPC error with no :id. Because the client correlates responses to pending requests by :id, that error was routed to the notification handler (and dropped), so the in-flight tool call's :on-error never fired and the call blocked until timeout — even though the HTTP status was known immediately.

Fixes:

- Bump plumcp to 0.2.2, which stamps the originating request's :id onto synthesized HTTP errors. The error now reaches the pending call's :on-error promptly with :plumcp.core/http-status, so the existing reinit-worthy-http-status? path flags the server for re-initialization (token refresh, or browser re-auth when the refresh token is also dead) instead of hanging.

- Proactively refresh a locally-expired OAuth token before issuing a tool call, so a known-expired token skips the doomed request and full reinit entirely.

- [x] This is not an AI slop.


ericdallo (approved):
Thank you!

PR #515, opened 2026-06-25
Fix Vanishing Chats and Messages from Remote

MrnoRac:
In the remote web UI two bugs appear for chats whose stored `:id` field has drifted from their `:chats` map key:

1. **Chat vanishes from the list** — it's missing from the sidebar and disappears again on reload, even though it's fully loaded with messages and reachable via `GET /api/v1/chats/<id>`.
2. **Chat loses all messages when switching tabs** — first open shows everything, but switching away and back renders it empty 

Reproducible on versions:

- eca (server): master @ `efd534ea6` (ECA_VERSION `0.142.2`)
- eca-web: `543b428` (eca-webview submodule `1790003`)
- eca-emacs: `20260622.1430` (also reproduced on `20260608.2031`)

Setup:** an ECA server with `remote.enabled: true`, an editor session (e.g. eca-emacs) with at least two pre-existing chats that have message history, and the web UI (`web.eca.dev` or a local `eca-web`) connected to it.

**Bug 1 — chat vanishes from the list / on reload:**
1. In the editor, `/resume` (or `/sessions`) an existing chat with history.
2. In the web UI it appears live in the sidebar.
3. **Reload the web page** → the resumed chat is **gone** from the sidebar, even though `GET /api/v1/chats/<id>` still returns it with all its messages.

**Bug 2 — chat loses messages on tab switch:**
1. In the web UI, open chat 1 → its messages render.
2. Open chat 2 → its messages render.
3. Switch back to chat 1 → **empty** (no messages). Switch back to chat 2 → **also empty**.
4. Switching to a different connection/session and back makes the *first* open of each chat show its messages again — then it breaks the same way on the next switch.

## Proposed Fix

Treat the map key as the authoritative id everywhere in the remote handlers as is specified here:
`src/eca/features/chat.clj:1967`:

> "The map key is used as the authoritative `:id` because legacy DB rows may have been persisted without an `:id` inside the value, and clients need a non-nil id to call `chat/open`."

`src/eca/db.clj:326`:

> "Ensures every chat value carries its map key as `:id`, so readers can rely on it. Heals legacy rows persisted before chats were seeded with an `:id`."

So...

- `handle-list-chats` / `session-state`: iterate `:chats` pairs, filter `editor-open` by key, report the key as `:id`
- `handle-get-chat`: return the requested path-param key as `:id`

Cannot affect forks (they already have `:id == key`).

## Tests

Added regression tests in `handlers_test.clj`: a chat whose `:id` field ≠ its key, marked editor-open, must appear in `handle-list-chats` (reported by key) and round-trip through `handle-get-chat` with the key. `bb test` green.

## Note / follow-up

This is just in case drift happens somewhere else, but it's probably a good idea to fix that drift there too.. it seems to happen in eca-emacs on /resume for example.

- [x] This is not an AI slop.
- [x] Messed around with the checklist :)


MrnoRac:
I tried to follow the patterns I saw around the repo, but let me know if anything needs changing.

ericdallo (approved):
thank you!

PR #524, opened 2026-07-10
Discover GitHub Copilot model capabilities

zikajk:
Use authenticated /models metadata to select Chat, Responses, or Messages APIs and expose only provider-advertised reasoning variants. Preserve user overrides and propagate discovered capabilities through model aliases, the variant picker, subagents, and cross-API history handling.

Support adaptive and budget-based Claude thinking through Copilot's Messages API without guessing variants when discovery metadata is unavailable.

- [x] This is not an AI slop.


zikajk:
@ericdallo Please, take a look.
I haven't created any fallback when `/models` is not accessible. I think we don't need it.

ericdallo (approved):
cool

beetleman:
@ericdallo I do not test it but this is look fix for this issue: https://github.com/editor-code-assistant/eca/issues/526

PR #529, opened 2026-07-13
Add configurable parent requirements for subagents

rschmukler:
Add spawnableBy discovery filtering and server-side authorization so workflow-specific subagents stay private while unrestricted agents retain existing behavior.

- [x] This is not an AI slop.

### DIFF
--- a/src/eca/features/agents.clj
+++ b/src/eca/features/agents.clj
@@ -60,12 +60,40 @@
     (sequential? tools) {"byDefault" "ask" "allow" (vec tools)}
     :else nil))
 
+(defn ^:private normalize-agent-id
+  [agent-id]
+  (when (string? agent-id)
+    (let [normalized (-> agent-id string/trim string/lower-case)]
+      (when-not (string/blank? normalized)
+        normalized))))
+
+(defn ^:private normalize-spawnable-by
+  [spawnable-by]
+  (cond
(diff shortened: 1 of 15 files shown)


src/eca/features/agents.clj
```diff
@@ -60,12 +60,43 @@
...
+
+(defn ^:private normalize-spawnable-by
+  [spawnable-by]
+  (cond
+    (nil? spawnable-by) nil
+    (string? spawnable-by) (or (normalize-agent-id spawnable-by)
+                               (do
+                                 (logger/warn logger-tag "Ignoring blank spawnableBy agent id")
+                                 nil))
+    (coll? spawnable-by) (let [normalized (->> spawnable-by
```
zikajk: This can be rewritten as this (you can get rid of let binding etc.)
```
(some->> spawnable-by
         (keep (fn [agent-id]
				    (or (normalize-agent-id agent-id)
					    (logger/warn logger-tag (format "Ignoring malformed spawnableBy agent id: %s" (pr-str agent-id))))))
         distinct
         vec
	     not-empty)
```

zikajk (dismissed):
Just one minor point, but otherwise everything looks good. Thank you.

ericdallo (approved):
Thanks

PR #531, opened 2026-07-14
Fix Codex OAuth rate limit reset handling

itkonen:
Since a recent upgrade the, Codex sessions have started to stall randomly after finishing a step, for example a tool call.

There seems to be a bug in the way rate limits are read from the messages, and by default there's cap on the wait time. This PR should fix it.

# AI Summary

## Problem

ECA could put a ChatGPT/Codex OAuth session to sleep for hours after a 429 response. It inferred retry timing from apparently exhausted `x-codex-primary-*` and `x-codex-secondary-*` quota-window headers and selected the latest reset, while ignoring the authoritative reset in the response body. Because `rateLimitMaxWaitSeconds` was optional, that inferred delay was uncapped by default.

The behavior was checked against the official `openai/codex` implementation at commit `f90e7deea6a715bbd153044af6f475eefa749177`:

- `codex-rs/codex-api/src/api_bridge.rs` handles a 429 `usage_limit_reached` response by reading numeric `error.resets_at` Unix epoch seconds from the JSON body.
- The `x-codex-*` primary and secondary window fields are quota snapshots; they are not independently used as retry delays.
- A generic 429 without `usage_limit_reached` follows the normal bounded retry path.

## Fix

- Read Codex OAuth `error.resets_at` only when `error.type` is `usage_limit_reached`.
- Accept raw JSON bodies and already-decoded maps with string or keyword keys.
- Keep `retry-after` as the highest-priority signal, while allowing expired higher-priority values to fall through to a usable future reset.
- Stop treating Codex quota-window snapshot headers as retry instructions.
- Preserve existing Anthropic and direct OpenAI reset-header handling.
- Apply a default 60-second `rateLimitMaxWaitSeconds` cap to the total scheduled delay, including the existing one-second safety buffer; provider overrides remain supported.
- Surface the provider reset timestamp immediately when the structured wait exceeds the cap.
- Update the config schema and Unreleased changelog entry.

## Verification

- `clojure -M:test --focus eca.llm-providers.errors-test` — 9 tests, 83 assertions
- `clojure -M:test --focus eca.llm-api-test` — 24 tests, 323 assertions
- `clojure -M:test --focus eca.llm-providers.openai-test` — 21 tests, 75 assertions
- `LC_ALL=C bb test` — 778 tests, 4,274 assertions, 0 failures
- `python3 -m json.tool docs/config.json` — passed
- `git diff --check` — passed
- Tested manually with the local debug server and a Codex OAuth session.

`clj-kondo` was not available in the local environment.

- [x] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix ChatGPT/Codex OAuth rate-limit retries using the response reset instead of quota snapshot headers; cap waits at 60 seconds by default.
 - Fix prompt cache invalidation warning after clearing the chat and changing model. #530
 - Fix missing line break after "Prompt stopped" message when followed by another system message.
 
--- a/docs/config.json
+++ b/docs/config.json
@@ -696,8 +696,10 @@
         },
         "rateLimitMaxWaitSeconds": {
           "type": "integer",
-          "description": "Optional cap on how long to wait for a rate limit reset (derived from provider response headers like retry-after). Unset: ECA sleeps until the reset time and auto-resumes, showing the reset time in chat. When set and the reset is further away, ECA gives up retrying and surfaces the reset time in the error.",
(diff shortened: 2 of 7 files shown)


ericdallo (approved):
Thank you!

PR #532, opened 2026-07-15
Retry transient OpenAI Responses failures

itkonen:
Codex sessions can get interrupted by `response.failed` events while streaming, which currently just surface an error without attempting a retry. This PR routes those failure events through ECA's existing retry flow, so the session can continue when the error is transient and no partial response has already been shown.

## AI Summary

> This PR adds structured handling for OpenAI Responses API failures in both streaming and synchronous requests. When a stream emits `response.failed`, ECA now stops consuming the stream immediately, closes the HTTP reader, and only then forwards the failure to the existing retry orchestration. This also prevents stale events appearing after the failure from emitting text, preparing tool calls, or completing the response.
>
> The full provider error payload is preserved for diagnostics and classification. ECA also records the response ID, normalized request ID, and response headers when available. HTTP 200 responses made with `stream: false` now recognize a failed response body and produce the same structured error shape as the streaming path.
>
> Retry classification remains deliberately conservative. Structured `server_error` failures are treated as overloaded, while `rate_limit_exceeded` failures use the existing rate-limit handling. The generic transient message observed from Codex is accepted only as a fallback when OpenAI did not provide a structured error code or type. Unknown structured failures remain non-retryable, and user-defined retry rules continue to take precedence.
>
> Retries reuse ECA's existing bounded retry implementation, including exponential backoff, rate-limit reset headers, retry limits, cancellation checks, and retry notifications. No separate provider-specific retry loop was introduced.
>
> To avoid duplicating content, automatic retries are allowed only before user-visible output has been emitted. Text, reasoning, tool-call preparation, web-search activity, and image-generation activity all disable further retries for that response. Synchronous responses remain safe to retry because their output is buffered until the request succeeds.
>
> Regression tests cover structured streaming and non-streaming failures, stream cleanup and trailing-event suppression, request and response metadata, transient and non-transient classification, custom retry-rule precedence, asynchronous and synchronous retry success, and prevention of retries after partial output.

---

- [x] This is not an AI slop.

🤖 Generated with [ECA](https://eca.dev) (openai/gpt-5.6-sol - max)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Retry transient OpenAI Responses `response.failed` server errors before output, preserving structured error and request IDs.
 - Fix prompt cache invalidation warning after clearing the chat and changing model. #530
 - Fix missing line break after "Prompt stopped" message when followed by another system message.
 
--- a/src/eca/llm_api.clj
+++ b/src/eca/llm_api.clj
@@ -465,8 +465,7 @@
                                            (catch Exception e
                                              (logger/warn logger-tag "on-history-sanitized callback failed" {:exception (ex-message e)})))))
         emit-first-message-fn (fn [& args]
-                                (when-not @first-response-received*
-                                  (reset! first-response-received* true)
+                                (when (compare-and-set! first-response-received* false true)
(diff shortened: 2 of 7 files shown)


ericdallo (approved):
Thank you!

PR #533, opened 2026-07-15
Scope restored chat selections per chat

rschmukler:
Include chat-scoped model, variant, and trust updates during open, resume, and import, and return an atomic selection snapshot from chat/open.

## Summary

- Scope model, variant, and trust restoration from `chat/open`, `/resume`, and `/import` to the affected chat using `chatId`.
- Return an atomic `selection` snapshot from `chat/open` containing the effective model, agent, variant, available variants, and trust state.
- Keep scoped `config/updated` notifications for backward compatibility.
- Correct the `chat/open` response field from `found?` to the documented `found`.
- Add unit and end-to-end regression coverage for selection restoration.

## Why

Previously, restoring chat A emitted session-wide `config/updated` notifications. Clients could therefore apply chat A’s model, variant, or trust state to chat B and to the defaults inherited by future chats.

Scoping these updates isolates each chat’s selection state. Including the complete selection in the `chat/open` response also lets clients hydrate a restored chat atomically rather than reconstructing its state from multiple independently ordered notifications.

- [x] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Fix prompt cache invalidation warning after clearing the chat and changing model. #530
 - Fix missing line break after "Prompt stopped" message when followed by another system message.
+- Scope restored model, variant, and trust selections to the opened, resumed, or imported chat instead of changing other chats and session defaults.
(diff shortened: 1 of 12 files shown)


ericdallo (approved):
Server looks good, thanks!

PR #537, opened 2026-07-16
fix: truncate oversized failed tool output

rschmukler:
- [x] This is not an AI slop.


ericdallo (approved):
good catch!

PR #542, opened 2026-07-19
docs: add Cerebras provider  example

### LINKED ISSUE #539: Support Cerebras as provider
I've been playing around with [Cerebras](https://www.cerebras.ai/inference) and the speed is just insane.. they have Gemma and GLM that are pretty decent models for "live" interactive coding, the kind that works great with ECA (as opposed to hands off agent loop style coding)

Could we add them as a provider? Their API is OpenAI compatible.

https://inference-docs.cerebras.ai/api-reference/chat-completions
https://inference-docs.cerebras.ai/resources/openai

Ramblurr:
This PR adds an example to the documentation for using the Cerebras provider which is OpenAI compatible. It is very very fast, 1k + tokens per second.

- [x] This is not an AI slop.

fixes #539

### DIFF
--- a/docs/config/models.md
+++ b/docs/config/models.md
@@ -59,6 +59,39 @@ ECA support lots of providers via its supported APIs (openai-chat, openai-respon
     }
     ```
 
+=== "Cerebras"
+
+    [Cerebras](https://www.cerebras.ai/inference) provides an OpenAI-compatible Chat Completions API.
+
+    Export your API key:
+
+    ```bash
+    export CEREBRAS_API_KEY="csk-..."
+    ```
+
+    Configure Cerebras as a custom `openai-chat` provider:
+
+    ```javascript title="~/.config/eca/config.json"
+    {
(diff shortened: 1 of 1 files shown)


ericdallo (changes requested):
Actually, reviewing this, I think this is already possible via custom providers, so no need for code for that right?
just do the same we recomend here, like [deepseek for example](https://eca.dev/config/models/#deepseek)
This makes easier to maintain ECA

Ramblurr:
@ericdallo Yea, fair enough! I've changed this PR to be a simple docs update.

One thing I noticed, however, is that it's not possible to specify the modalities using config. For example this provider has some models capable of vision and others that aren't. Something for another issue perhaps?

ericdallo:
@Ramblurr yes, we can discuss in other issue, but we do have a concept of image support from capabilities that we take from https://models.dev

ericdallo (approved):

PR #553, opened 2026-07-23
Re-emit chat config after plugin resolution

rschmukler:
Plugin-provided agents (and anything from installed plugins) can be missing from
the editor's agent picker, even though the server has loaded them correctly.
Symptom: the editor shows only built-in agents (`code`, `plan`) plus any
project-local `.eca/agents/*` agents, while plugin MCP servers still show up as
available. Querying the server's `/api/v1/session` shows the full agent list, so
the agents are loaded — the editor just never gets told about them.

## Root cause

At `initialized`, two independent futures run:

1. **Model sync** — on completion it sends a `config/updated` notification that
   carries the agent list (`:agents (primary-agent-names …)`).
2. **Plugin resolution** — `reset!`s `plugin-components*`, which is what makes
   plugin agents visible to `config/all`.

These race. When model sync finishes **before** plugin resolution, it advertises
the pre-plugin agent list, and nothing re-emits it afterward — so plugin agents
never reach the editor. This is reliably hit with providers configured with
`"fetchModels": false`, since model sync then returns almost instantly and
always wins the race. (It also explains `selectAgent` falling back to `code`
when the configured `defaultAgent` is a plugin agent.)

## Fix

- Extract the `config/updated` payload building into `notify-chat-config-state!`.
- The model-sync callback calls it (unchanged behavior — it already re-reads
  `config/all`).
- Call it again right after `config/deliver-plugins-resolved!` in the plugin
  future.

`notify-fields-changed-only!` diffs against the session-level mirror, so the
extra call only emits the newly-resolved agents and is a no-op when plugins won
the race. Either ordering now converges on the full agent list.

## Testing

- `clj-kondo` clean (0 errors / 0 warnings).
- Namespace compiles.
- `eca.handlers-test` (12 tests / 78 assertions) and
  `eca.features.plugins-test` + `eca.config-test` + `eca.server-test`
  (32 tests / 260 assertions) pass.

- [x] This is not an AI slop.


ericdallo (approved):
thank you!

PR #556, opened 2026-07-24
Return task state after each operation

AndiLavera:
Changes `task.clj` to return the task state after each operation, explicitly telling the LLM the state after an operation. This may help smaller models 'remember' the latest state, or prevent a second turn for reading the state after mutating for more eager models.

- [x] This is not an AI slop.


ericdallo:
thanks!

PR #559, opened 2026-07-27
Fix OpenAI Responses chats stuck after SSE error events

rschmukler:
## Summary

Handle terminal SSE `error` events from the OpenAI Responses API and route them through ECA's existing error and retry handling.

## Problem

An OpenAI Responses request can return HTTP 200 but terminate with an SSE error instead of `response.completed`, for example:

- Event: `error`
- Code: `server_is_overloaded`

ECA previously handled `response.failed`, but generic SSE `error` events were passed to the stream callback and ignored. When the stream then closed normally, ECA emitted neither an error nor a finish event, leaving the chat marked as active indefinitely.

## Changes

- Treat SSE `error` events as terminal OpenAI Responses errors.
- Preserve structured error fields, response headers, and request IDs.
- Close the response stream before invoking error or retry handling.
- Classify `server_is_overloaded` as a retryable provider overload.
- Integrate the handling with the existing stream cancellation and idle-timeout watchdog.
- Add regression tests covering:
  - HTTP 200 streams ending with an SSE `error`.
  - Ignoring trailing events after the terminal error.
  - Stream closure before error handling.
  - `server_is_overloaded` classification.
  - Compatibility with stream cancellation handling.

## Verification

- 35 tests passed.
- 197 assertions passed.
- No test failures.
- No new `clj-kondo` errors.
- Editor diagnostics are clean.

- [x] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix OpenAI Responses SSE errors, incomplete responses, and premature stream closes so chats retry or finish instead of staying active.
+
 ## 0.150.1
 
 - Add Claude Opus 5 support.
--- a/src/eca/llm_providers/errors.clj
+++ b/src/eca/llm_providers/errors.clj
@@ -51,6 +51,10 @@
    #"(?i)connection refused"
    #"(?i)UnresolvedAddressException"])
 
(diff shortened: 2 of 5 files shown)


ericdallo (approved):
Makes sense, thanks!

PR #565, opened 2026-07-31
Keep compact tool schema stable accross compaction

zikajk:
- [x] This is not an AI slop.


zikajk:
@ericdallo This is what we've discussed, looks good based on quick testing (LLM can call compact anytime but it will be rejected).

ericdallo (approved):
Cool, thanks!

PR #578, opened 2026-08-22
Restore parallel tool calls for Codex OAuth

### LINKED ISSUE #572: Support parallel tool execution for Codex OAuth models

Users accessing recent Codex models through OpenAI OAuth/ChatGPT subscriptions currently get serial tool execution in ECA. Independent operations such as reading several files require a separate model continuation for each tool call.

The tools often finish in milliseconds, while every model continuation can take several seconds. This makes ECA noticeably slower and may consume more tokens or subscription quota than executing a batch concurrently. Subagents using the same models are affected as well.

ECA's executor already runs tool-call batches concurrently when a provider supplies several calls at once. The limitation occurs earlier: `gpt-5.6-sol`, `gpt-5.6-terra`, and `gpt-5.6-luna` use Codex Responses Lite, which requires:

```clojure
{:parallel_tool_calls false}
```

Sending `true` returns HTTP 400:

```text
X-OpenAI-Internal-Codex-Responses-Lite requires
`parallel_tool_calls` to be false.
```

The official Codex CLI nevertheless performs concurrent work with OAuth accounts. It also sends `parallel_tool_calls=false`, but current models use Codex Code Mode: one outer `exec` call can invoke several nested tools concurrently.

This may be a broader provider-design problem rather than an OpenAI-specific request flag. Supporting composite or nested tool calls could affect provider adapters, approvals, hooks, cancellation, UI state, and persistent chat history.

A naive solution would be an ECA-specific batch function containing an array of tool calls. ECA could expand it through the existing concurrent executor and aggregate the results into one provider-visible output. However, this risks adding provider-specific behavior and requires decisions about history, partial failures, approvals, and model reliability. Full Codex Code Mode compatibility would be cleaner parity but substantially larger in scope.

It would be useful to decide on an ECA-level design for enabling concurrent tool execution when a provider cannot express it through ordinary parallel function-call batches.

itkonen:
Getting parallel tool calls work with Codex with OAuth proved much simpler than I feared. Oddly the API requires the "Responses Lite" API with codex-5.6 models, but then it advertises parallel support allowing to use the full Responses API when returning parallel tool calls. I bet it wasn't designed by humans. 

## Summary
- preserve the explicit live combination of Responses Lite and parallel tool-call support, then use the ordinary Responses function-tool shape when batching is requested
- retain `all_turns` reasoning context and function-tool visibility without enabling hosted web or image tools
- keep existing Lite behavior for fallback, partial, disabled, and no-function-tool cases

## Testing
- `clojure -M:test --focus eca.llm-providers.openai-test`
- `clojure -M:test --focus eca.llm-api-test`
- `clojure -M:test --focus eca.features.chat-test/concurrent-tool-calls-test`
- `LC_ALL=C LANG=C bb test`
- `clj-kondo --lint src/eca/llm_providers/openai.clj test/eca/llm_providers/openai_test.clj`

Closes #572

🤖 Generated with [ECA](https://eca.dev) (openai/gpt-5.6-sol - xhigh)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Restore parallel function-call batching for ChatGPT OAuth models that advertise Responses Lite and parallel tool-call support. #572
+
 ## 0.155.1
 
 - Fix Anthropic streaming dropping tool calls when providers report `end_turn`/`stop` stop_reason with pending `tool_use` blocks. #576
--- a/src/eca/llm_providers/openai.clj
+++ b/src/eca/llm_providers/openai.clj
@@ -130,27 +130,34 @@
      :discovered-variants (codex-reasoning-variants supported-reasoning-efforts)}))
 
 (defn ^:private codex-live-model-discovery
-  "Maps a Codex /models entry into generic discovery keys. Codex-only model
-   behavior goes inside :discovered-provider-data, interpreted only here."
+  "Maps a Codex /models entry into provider-specific discovery metadata."
   [{:keys [use_responses_lite default_reasoning_level
            supported_reasoning_levels supports_parallel_tool_calls] :as model}]
   (let [efforts (codex-normalize-reasoning-efforts supported_reasoning_levels)
         provider-data (assoc-some
                        (cond-> {}
                          (contains? model :use_responses_lite)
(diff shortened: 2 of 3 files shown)


ericdallo (approved):
Looks good, LMK if any issues with that, thank you

PR #585, opened 2026-09-06
fix: honor no_proxy for environment-configured HTTP proxies

### LINKED ISSUE #583: Add support for no_proxy variable along with http_proxy

Please avoid AI slops, be concise, and focus on what matters for this issue.

**Describe the bug**

The server does load the `http_proxy` variable, but does not load `no_proxy`, hence the connection to internal gateway API are not possible. 

**To Reproduce**

1. Start ECA server with environment variables `https_proxy`, `http_proxy`, `no_proxy` set system-wide
2. Start eca client
3. /model
4. `403: blocked by proxy`

**Expected behavior**

Connect to the internal API gateway.

**Additional context**
Variable which is required in certain condition notably when you are behind a proxy, thus require the `http_proxy`, but your API gateway is internal (home network for example) and must not go through the proxy. 

I can see that there is no code related to the variable no_proxy in https://github.com/editor-code-assistant/eca/blob/master/src/eca/network.clj

emecii:
- [x] This is not an AI slop.

Fixes #583.

ECA currently sends internal model/API requests through its configured HTTP proxy even when `no_proxy` names the destination. Read `no_proxy` (falling back to `NO_PROXY`) and select a direct route for matching hosts in the existing Hato HTTP/HTTPS proxy selector.

The bypass list supports case-insensitive host/domain matching, dot-boundary subdomains, exact IP literals, and a standalone `*`. Lowercase takes precedence, including an empty value. Documentation specifies that CIDR ranges, port-qualified entries, and partial wildcards are unsupported. Other HTTP clients and JVM-global proxy settings are outside this change.

Validation:

- Both network/client namespaces pass: 14 tests, 122 assertions. The new regressions fail on the original source with 26 assertion failures.
- Real loopback HTTP servers verify direct versus proxy routing through Hato. Selector tests cover HTTPS; successful end-to-end TLS bypass was not tested.
- Changed-file clj-kondo and formatting pass. Repository-wide lint reports one error and 16 warnings, identical on the base revision (the error is the existing unresolved `dir` in `hooks_test.clj`).
- Full integration suites and native-image/platform builds were not run locally.

AI assistance: OpenAI Codex (GPT-6) was used for investigation, implementation, review, and test execution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Honor `no_proxy`/`NO_PROXY` when routing HTTP and HTTPS requests through environment-configured proxies. #583
+
 - Add `gpt-6-astra` support: built-in variants `low`, `medium`, `high`, `xhigh`, `max` (no `none`), Codex Lite/context fallbacks, web search and image generation.
 
 ## 0.158.1
--- a/docs/config/network.md
+++ b/docs/config/network.md
@@ -21,6 +21,30 @@ https_proxy="http://user:pass@host:port"
 
 Lowercase wins if both are set. Credentials (if used) must match for HTTP and HTTPS.
 
+### Bypassing the proxy
+
+For ECA's Hato HTTP client (including LLM API requests), set `no_proxy` or `NO_PROXY`:
+
+```bash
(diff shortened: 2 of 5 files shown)


ericdallo:
Thank you!

PR #590, opened 2026-09-10
Support @file context in custom command arguments

### LINKED ISSUE #163: Feature: Support file content (@) as custom-command arg
I want to be able to pass # or @ as an argument for custom commands. 

It looks to me that `#` already works, but `@` doesn't seem to add the content of the file. 

emecii:
## Summary

- resolve `@file` custom-command and parameterized-skill arguments after substitution and attach their contents through the existing prompt-context format
- share path/range parsing with regular prompt contexts, including `~` expansion and trailing ranges on Windows-style paths
- preserve quoted arguments containing spaces and document the quoting requirement

Fixes #163

## Testing

- `clojure -M:test --focus eca.features.commands-test` — 27 tests, 199 assertions
- `clojure -M:test --focus eca.features.context-test` — 11 tests, 54 assertions
- `clojure -M:test --focus eca.features.chat-test` — 72 tests, 468 assertions
- `bb test` — 934 tests, 5,374 assertions
- `bb integration-test --dev --ns integration.chat.commands-test` — 4 tests, 74 assertions
- `clj-kondo` on changed source/tests — 0 errors, 0 warnings
- Whole-tree `clj-kondo` only reports the pre-existing unresolved `dir` in `test/eca/features/hooks_test.clj:259`

This contribution was prepared with OpenAI Codex assistance.

- [x] Added an Unreleased changelog entry.

### DIFF
--- a/docs/config/commands.md
+++ b/docs/config/commands.md
@@ -7,7 +7,7 @@ description: "Configure ECA commands: built-in slash commands like /init and /co
 ![](../images/features/commands.png)
 
 You can configure custom command prompts for project, global or via `commands` config pointing to the path of the commands.
-Prompts can use positional variables like `$ARGUMENTS`, `$1`, `$2`, or [named `{{name}}` variables](#frontmatter-description-and-named-arguments), to replace in the prompt during command call.
+Prompts can use positional variables like `$ARGUMENTS`, `$1`, `$2`, or [named `{{name}}` variables](#frontmatter-description-and-named-arguments), to replace in the prompt during command call. File references passed as arguments (for example, `@/workspace/src/app.clj`) include the file content just like they do in a regular chat prompt. Quote arguments that contain spaces.
 
 !!! tip "Skills support arguments too"
 
--- a/src/eca/features/commands.clj
+++ b/src/eca/features/commands.clj
@@ -10,6 +10,7 @@
    [eca.features.chat.debug :as f.chat.debug]
    [eca.features.chat.export :as f.chat.export]
    [eca.features.chat.lifecycle :as lifecycle]
+   [eca.features.context :as f.context]
    [eca.features.index :as f.index]
    [eca.features.login :as f.login]
    [eca.features.plugins :as f.plugins]
@@ -338,6 +339,20 @@
   (when-let [raw-content (:content (first (filter #(= command (:name %))
                                                   custom-cmds)))]
     (substitute-args raw-content args)))
+
+(defn ^:private file-context-arg [arg]
+  (when (and (string/starts-with? arg "@")
+             (< 1 (count arg)))
+    (f.context/parse-context-reference (subs arg 1))))
+
+(defn ^:private add-file-context [prompt args db]
+  (if-let [contexts-str (some-> (keep file-context-arg args)
+                                (f.context/raw-contexts->refined db)
+                                seq
+                                (f.prompt/contexts-str nil nil))]
+    [{:type :text :text prompt}
+     {:type :text :text contexts-str}]
+    prompt))
 
 (defn ^:private format-tool-permissions [{:keys [toolCall]}]
   (when-let [approval (:approval toolCall)]
@@ -1152,7 +1167,7 @@
       ;; else check if a custom command or skill
       (if-let [custom-command-prompt (get-custom-command command args custom-cmds)]
         {:type :send-prompt
-         :prompt custom-command-prompt}
+         :prompt (add-file-context custom-command-prompt args db)}
         (if-let [skill (first (filter #(= command (:name %)) skills))]
           {:type :send-prompt
            :prompt (cond
@@ -1163,7 +1178,7 @@
                      (str "Load skill: " (:name skill))
 
                      (seq args)
-                     (substitute-args (:body skill) args)
+                     (add-file-context (substitute-args (:body skill) args) args db)
 
                      :else
                      (str "Load skill: " (:name skill)))}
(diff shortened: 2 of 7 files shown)


src/eca/features/commands.clj
```diff
@@ -339,6 +340,26 @@
                                                   custom-cmds)))]
     (substitute-args raw-content args)))
 
+(defn ^:private file-context-arg [arg]
+  (when (and (string/starts-with? arg "@")
+             (< 1 (count arg)))
+    (let [reference (subs arg 1)
+          [_ path start end] (re-matches #"(.+):L(\d+)-L(\d+)$" reference)
```
zikajk: This adds a second parser for the `@path(:L1-L4)` syntax. `eca.features.context/contexts-str-from-prompt` already parses the same syntax out of prompt text, and the two regexes disagree:

there: `@"@([/~\.][^\s:]+)(?::L(\d+)-L(\d+))?"`. The path must start with `/`, `~` or `.`, and matching stops at the first space or colon.
here: `@"(.+):L(\d+)-L(\d+)$"` applied to the already-tokenized arg. This is the better definition: quoted args keep spaces, and the trailing range split handles Windows paths, so `@C:\dir\f.clj:L2-L4` parses here but not in the chat scanner (both cases checked in a REPL).

Please make the working definition the single source of truth: extract a helper in `eca.features.context`, for example:

 ```clojure
 (defn parse-context-reference
   "Parses `<path>` or `<path>:L<start>-L<end>` into a raw context map
    ({:type :file|:directory :path ... :lines-range ...}), or nil."
   [^String reference]
   ...)
 ```

- `file-context-arg` then becomes a thin `@` prefix check plus a call to the helper.
- `contexts-str-from-prompt` keeps its own token-finding regex, because its acceptance rules must stay: the `/`, `~`, `.` prefix and the no-space limit are load-bearing (there is a test in `test/eca/features/context_test.clj` asserting that emails like `user@example.com` do not match). Only the range split and the raw-map building move into the helper.

One extra fix that fits naturally in the same helper: nothing expands `~` today, so `@~/notes.md` fails in both flows (`fs/readable?` sees a literal `~` directory). A single `fs/expand-home` there fixes chat mentions and command args at once.

src/eca/features/commands.clj
```diff
@@ -1152,7 +1173,7 @@
       ;; else check if a custom command or skill
       (if-let [custom-command-prompt (get-custom-command command args custom-cmds)]
         {:type :send-prompt
-         :prompt custom-command-prompt}
+         :prompt (add-file-context custom-command-prompt args db)}
```
zikajk, line 1170: NITPICK: 
Skills with args take the same route but get no file context. A few lines below, the skill branch (`(seq args) (substitute-args (:body skill) args)`) also produces a `:send-prompt` built from the same `args`, so after this PR `/my-cmd @/abs/file.clj` attaches the file content while `/my-skill @/abs/file.clj` attaches nothing. The substitution mechanics are identical, so wrapping that branch with `add-file-context` is the same one-liner and keeps custom commands and parameterized skills consistent. Can you add it there as well and update the changelog?

docs/config/commands.md
```diff
@@ -7,7 +7,7 @@ description: "Configure ECA commands: built-in slash commands like /init and /co
 ![](../images/features/commands.png)
 
 You can configure custom command prompts for project, global or via `commands` config pointing to the path of the commands.
-Prompts can use positional variables like `$ARGUMENTS`, `$1`, `$2`, or [named `{{name}}` variables](#frontmatter-description-and-named-arguments), to replace in the prompt during command call.
+Prompts can use positional variables like `$ARGUMENTS`, `$1`, `$2`, or [named `{{name}}` variables](#frontmatter-description-and-named-arguments), to replace in the prompt during command call. File references passed as arguments (for example, `@/workspace/src/app.clj`) include the file content just like they do in a regular chat prompt.
 
 !!! tip "Skills support arguments too"
```
zikajk, line 13: NITPICK: This sentence implies full parity with regular chat mentions, and that only holds for the failure case. With a space in the path:

- Unquoted (`/review @/dir/My File.clj`): `tokenize-args` splits on whitespace, so the server sees args `@/dir/My` and `File.clj`. Same truncation happens for the same path typed in a normal prompt, where the mention regex `[^\s:]+` stops at the space. Both drop the file.
- Quoted (`/review "@/dir/My File.clj"`): works with this PR, and this is the only way to pass a space path. Note this rescue is command-specific: in a normal prompt a quoted mention does not match the mention regex at all (`@` must be followed by `/`, `~` or `.`), so the equivalent chat input attaches nothing.

The new test passes because it builds the args vector directly and bypasses `tokenize-args`; worth adding one assertion through `tokenize-args` for the quoted form. Please also reword the doc line so the quoting requirement is stated, for example: "File references passed as arguments (for example, `@/workspace/src/app.clj`) include the file content just like they do in a regular chat prompt. Quote arguments that contain spaces."

zikajk:
@emecii Nice feature, and the happy path is solid! Just one blocker.

zikajk (changes requested):

src/eca/features/commands.clj
emecii: Extracted parse-context-reference in eca.features.context and routed both prompt scanning and command args through it. The prompt regex keeps its existing prefix/no-space rules, while the helper owns trailing range parsing and home-directory expansion; tests cover ranges, Windows-style paths, home expansion, and email non-matching.

src/eca/features/commands.clj
emecii: Added the same add-file-context step to parameterized skills, plus focused coverage. The changelog now names both custom commands and parameterized skills.

docs/config/commands.md
emecii: Reworded the docs to require quoting arguments that contain spaces and added a tokenizer-level assertion showing the quoted file reference remains one argument.

zikajk (approved):
Looks good to me :1:

ericdallo (approved):
thank you!

PR #591, opened 2026-09-10
fix: retry Copilot token renewal server errors

### LINKED ISSUE #551: oauth-renew-token` crashes with full HTML body when GitHub returns 5xx Unicorn page

Description:**

When GitHub's `/copilot_internal/v2/token` endpoint returns a 5xx error (the "Unicorn" HTML error page), `oauth-renew-token` in `copilot.clj` does not check the HTTP status code before attempting to extract `:token` from the response body. Since `clj-http` with `:as :json` falls back to returning the raw string when the response isn't valid JSON, `(:token body)` evaluates to `nil`, and the function throws an `ex-info` containing the entire HTML page as the error message.

This causes two user-visible issues:

1. The log is flooded with a massive HTML dump instead of a concise error message.
2. There is no retry logic for transient GitHub failures — if the token renewal happens to hit a temporary 5xx during startup (`renew-expiring-auth-tokens!`) or at prompt time (`maybe-renew-auth-token`), the chat prompt fails immediately with `Auth token renew failed`.

**Steps to reproduce:**

1. Have a valid Copilot `access-token` stored in the auth cache.
2. Start ECA while GitHub's Copilot token endpoint is returning 5xx (or simulate by e.g. blocking the endpoint temporarily).
3. Observe the WARN log containing the full Unicorn HTML and the subsequent ERROR `Auth token renew failed`.

**Expected behavior:**

- `oauth-renew-token` should check the HTTP status before parsing the body.
- On 5xx or non-JSON responses, it should throw a concise error like `"GitHub Copilot token endpoint returned HTTP <status>"`.
- `renew-expiring-auth-tokens!` (and ideally `maybe-renew-auth-token`) should retry at least once with a short backoff for transient (5xx) errors before giving up.

**Relevant code:**

- `src/eca/llm_providers/copilot.clj` — `oauth-renew-token` (line ~66)
- `src/eca/features/login.clj` — `renew-auth!` / `renew-expiring-auth-tokens!`

**Logs:**

```
WARN  [LOGIN] Could not renew 'github-copilot' auth token before sync: Error on copilot login: <!DOCTYPE html>... [full Unicorn HTML page]
WARN  [MODELS] Provider 'github-copilot': /models endpoint returned status 401: IDE token expired
ERROR Auth token renew failed

### PR DESCRIPTION
**Pull request by emecii**
Fixes #551.

GitHub Copilot token renewal now retries one transient gateway/service response after a 250 ms delay. The retry allowlist is limited to HTTP 500, 502, 503, and 504; 511 and other statuses remain single-shot. Failed responses retain the original body in both the exception message and data for diagnosis.

Validation:
- `clojure -M:test --focus eca.llm-providers.copilot-test` — 13 tests, 77 assertions
- `bb test` — 934 tests, 5,374 assertions
- `clj-kondo` on changed source/tests — 0 errors, 0 warnings
- Whole-tree `clj-kondo` only reports the pre-existing unresolved `dir` in `test/eca/features/hooks_test.clj:259`

Implementation assistance: OpenAI Codex.

- [x] Added an Unreleased changelog entry.

### DIFF
--- a/src/eca/llm_providers/copilot.clj
+++ b/src/eca/llm_providers/copilot.clj
@@ -17,6 +17,12 @@
 
 (def ^:private default-client-id "Iv1.b507a08c87ecfe98")
 
+(def ^:private token-renew-max-attempts 2)
+
+(def ^:private token-renew-retry-delay-ms 250)
+
+(def ^:private token-renew-retry-statuses #{500 502 503 504})
+
 (defn ^:private github-base-url [provider-settings]
   (or (get-in provider-settings [:auth :url])
       "https://github.com"))
@@ -241,20 +247,29 @@
 
 (defn ^:private oauth-renew-token [provider-settings access-token]
   (let [token-url (str (github-api-base-url provider-settings) "/copilot_internal/v2/token")
-        {:keys [status body]} (http/get
-                               token-url
-                               {:headers (merge (auth-headers)
-                                                {"authorization" (str "token " access-token)})
-                                :throw-exceptions? false
-                                :http-client (client/merge-with-global-http-client {})
-                                :as :json})]
-    (if-let [token (:token body)]
-      (cond-> {:api-key token
-               :expires-at (:expires_at body)}
-        (get-in body [:endpoints :api]) (assoc :api-url (get-in body [:endpoints :api])))
-      (throw (ex-info (format "Error on copilot login: %s" body)
-                      {:status status
-                       :body body})))))
+        request-options {:headers (merge (auth-headers)
+                                         {"authorization" (str "token " access-token)})
+                         :throw-exceptions? false
+                         :http-client (client/merge-with-global-http-client {})
+                         :as :json}]
+    (loop [attempt 1]
+      (let [{:keys [status body]} (http/get token-url request-options)]
+        (cond
+          (:token body)
+          (cond-> {:api-key (:token body)
+                   :expires-at (:expires_at body)}
+            (get-in body [:endpoints :api]) (assoc :api-url (get-in body [:endpoints :api])))
+
+          (and (token-renew-retry-statuses status)
+               (< attempt token-renew-max-attempts))
+          (do
+            (Thread/sleep (long token-renew-retry-delay-ms))
+            (recur (inc attempt)))
+
+          :else
+          (throw (ex-info (format "Error on copilot login: %s" body)
+                          {:status status
+                           :body body})))))))
 
 ;; --- Settings-based login (providers/login flow) ---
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/eca/llm_providers/copilot.clj [hunk 1]
```diff
@@ -241,20 +245,34 @@
...
+        request-options {:headers (merge (auth-headers)
+                                         {"authorization" (str "token " access-token)})
+                         :throw-exceptions? false
+                         :http-client (client/merge-with-global-http-client {})
+                         :as :json}]
+    (loop [attempt 1]
+      (let [{:keys [status body]} (http/get token-url request-options)]
+        (cond
+          (and (number? status) (<= 200 status 299))
+          (if-let [token (when (map? body) (:token body))]
```
zikajk: There is no reason to check for `map` here, it is ok to just check the `:token` like it did before.

src/eca/llm_providers/copilot.clj
```diff
@@ -241,20 +245,34 @@
...
+            (cond-> {:api-key token
+                     :expires-at (:expires_at body)}
+              (get-in body [:endpoints :api]) (assoc :api-url (get-in body [:endpoints :api])))
+            (throw (ex-info "GitHub Copilot token endpoint returned an invalid response"
+                            {:status status})))
+
+          (and (number? status)
+               (<= 500 status 599)
```
zikajk: Does this really make a sense? E.g. I am not sure if it should retry on 511.

src/eca/llm_providers/copilot.clj
```diff
@@ -241,20 +245,34 @@
...
+               (< attempt token-renew-max-attempts))
+          (do
+            (Thread/sleep (long token-renew-retry-delay-ms))
+            (recur (inc attempt)))
```
zikajk, line 267: Are you sure this really helps? Have you experienced getting the error and fixing it by the retry mechanism?

src/eca/llm_providers/copilot.clj
```diff
@@ -241,20 +245,34 @@
...
+
+          :else
+          (throw (ex-info (if (number? status)
```
zikajk: Why do you think it is bad idea to get the full body in the log?

src/eca/llm_providers/copilot.clj
emecii: Removed the map guard. The success path now checks (:token body) directly.

src/eca/llm_providers/copilot.clj
emecii: Narrowed the retry allowlist to 500, 502, 503, and 504. A regression test now confirms that 511 makes exactly one request.

src/eca/llm_providers/copilot.clj
emecii: The evidence is the failure sequence reported in #551: a transient token-endpoint 5xx immediately aborts startup or prompt-time renewal. The policy remains deliberately small (one retry), and the deterministic 503→200 test verifies that this recovery path succeeds.

src/eca/llm_providers/copilot.clj
emecii: Agreed. Restored the original response body in both the exception message and ex-data; the exhausted-502 test asserts that diagnostic detail is preserved. The PR description and changelog now match.

zikajk (approved):
Looks good to me 👍

PR #592, opened 2026-09-10
Keep project plugin installs additive across config layers

zikajk:
## Summary
- Append `plugins.install` across config layers, preserving source definitions and later-plugin precedence. Other config arrays keep their current behavior.
- Add layer-local `plugins.installMode: "replace"` to preserve the previous exclusion behavior, including an empty list.
- Remove only global declarations on plugin uninstall, rather than reporting inherited plugins as removed. Include focused regression tests and migration documentation.

## Compatibility
This changes the default: global plugins previously hidden by a project install list will now load, including their hooks and tools. Users who relied on exclusion must set `installMode: "replace"` beside that list. Existing direct hook definitions are unchanged.

## Verification
- 69 tests, 630 assertions passed across config, plugins, commands, and handlers.
- Focused Clojure lint, editor diagnostics, and whitespace checks passed.
- Isolated from the separate hook-feedback work and unrelated local changes.

🤖 Generated with [ECA](https://eca.dev) (openai/gpt-6-astra - medium)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Recover Anthropic streaming responses interrupted by transient TLS `bad_record_mac` failures.
+- BREAKING: `plugins.install` now appends across config layers. Set `plugins.installMode` to `replace` beside the list to exclude inherited plugins as before.
 
 ## 0.159.0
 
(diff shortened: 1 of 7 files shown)


ericdallo (approved):
Yeah I was aware of this gap, this looks good, and the default makes sense

PR #597, opened 2026-09-10
feat: mcpToolSearch config

rschmukler:
Every tool sent to the LLM costs context on every request: its description plus its full input schema. With a few MCP servers connected that is thousands of tokens the model rarely needs.

Add `mcpToolSearch`, which trades that upfront cost for an extra round trip:

    "mcpToolSearch": {
      "includePattern": [".*"],
      "excludePattern": ["clojure-mcp"]
    }

Matching MCP tools are deferred - their schemas are withheld and only a compact catalog of names and truncated descriptions is rendered into the system prompt. When the model needs one it calls the new `eca__search_tools` tool, which ranks the deferred catalog against a query, returns the matches with their input schemas, and records them on the chat so they are sent as regular tools from then on.

Deferred tools are only withheld from the request payload; they stay resolvable and callable throughout, so a tool call that arrives before the search still executes.

Only MCP tools can be deferred. Native ECA tools are the agent's baseline capabilities, so a catch-all pattern never takes them away - use `disabledTools` to remove one of those. `eca__search_tools` is likewise never deferred, and is only offered to the model when at least one tool is actually deferred.

Pattern matching reuses the existing `disabledTools` engine rather than introducing a third dialect alongside it and the exact-match approval selectors: anchored Java regex against the builtin tool name or the `server__tool` full name, or an exact server name for all of its tools. That matcher is extracted as `tool-entry-matches?` and its regex compilation is now memoized, which also lets it warn once when an entry fails to compile - `*` is not a valid regex and was previously matched literally in silence.

Configurable globally, per agent, and in agent markdown frontmatter, where the object may be abbreviated to a list or string when only `includePattern` is needed:

    mcpToolSearch:
      - github__.*

The static prompt cache signature now tracks the deferrable set separately from the tool list, so changing the patterns mid-chat rebuilds the catalog.

- [x] This is not an AI slop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add `mcpToolSearch` config to keep MCP tool schemas out of context until the LLM loads them with the new `eca__search_tools` tool.
 - Recover Anthropic streaming responses interrupted by transient TLS `bad_record_mac` failures.
 - BREAKING: `plugins.install` now appends across config layers. Set `plugins.installMode` to `replace` beside the list to exclude inherited plugins as before.
 
--- a/src/eca/config.clj
+++ b/src/eca/config.clj
@@ -187,6 +187,7 @@
                                                  "eca__grep" {}
                                                  "eca__editor_diagnostics" {}
                                                  "eca__skill" {}
+                                                 "eca__search_tools" {}
                                                  "eca__task" {}
                                                  "eca__fetch_rule" {}
                                                  "eca__spawn_agent" {}}
@@ -210,6 +211,7 @@
                                                      "eca__grep" {}
                                                      "eca__editor_diagnostics" {}
                                                      "eca__skill" {}
+                                                     "eca__search_tools" {}
                                                      "eca__task" {}
                                                      "eca__fetch_rule" {}}
                                              :deny {"eca__shell_command"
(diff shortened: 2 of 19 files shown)


src/eca/llm_api.clj
```diff
@@ -247,7 +247,10 @@
            past-messages tools provider-auth sync? subagent? cancelled? prompt-cache-key]
     :or {on-error identity}}]
   (let [real-model (real-model-name model model-capabilities)
-        tools (when (:tools model-capabilities) tools)
+        ;; Deferred tools stay callable but their schemas are withheld until the
+        ;; model loads them with eca__search_tools.
+        tools (when (:tools model-capabilities)
```
ericdallo, line 252: I believe this will work only for the first request, when LLM enters the loop, and need to request again we will send all tools, cna you check that? worth a test for that too

ericdallo (changes requested):

src/eca/config.clj
```diff
@@ -238,6 +240,8 @@
    :skills []
    :extraConfigs []
    :disabledTools []
+   :mcpToolSearch {:includePattern []
```
ericdallo, line 243: should we defer all by default? I believe cursor and others are doing this no? as I believe it helps for most cases

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-- Tool prompts: push, git config, shell search/read and sub-agents run on explicit request; rule headings are tool-scoped.
+- Add `mcpToolSearch` config to keep MCP tool schemas out of context until the LLM loads them with the new `eca__search_tools` tool.
+- Fix deferred MCP tools being re-sent from the second request of a tool-call loop onward, undoing the context saving as soon as the model called any tool.
```
ericdallo: I think we just need one single line of changelog

rschmukler: Sorry about that!

ericdallo (approved):

PR #606, opened 2026-09-14
Fix remote answers to questions asked before a web client connects

itkonen:
## Why

I use ECA in Emacs and sometimes step away while it is working. When it pauses on `ask_user`, I want to open ECA Web on my phone, answer the question, and let the original session continue. I don't necessarily have the web app connected when the question appears.

Opening ECA Web afterward showed the paused chat, but not the interactive question and options. Investigating that exposed two separate gaps:

- ECA Web restored history, but not the pending question.
- The backend only registered remotely answerable questions when an SSE client was already connected.

The second issue meant that restoring the UI alone would not be enough. The chat API exposed the pending question and its original request ID, but posting an answer returned `404 question_not_found`. Connecting a web client afterward did not change that. Answering through the owning editor still worked.

## Backend change

This PR registers pending questions regardless of whether an SSE client is connected.

Editor and HTTP answers use the same atomic first-answer-wins path. Registration happens before invoking the editor, and request handles are attached safely even if an answer arrives immediately. Editor failures continue to propagate as errors rather than being reported as user cancellation.

The existing API and SSE payloads remain unchanged.

## Companion web change

ECA Web PR: https://github.com/editor-code-assistant/eca-web/pull/3

That PR restores pending questions through the existing question UI after loading the chat. It also prevents old cached snapshots from hiding a question when connecting or reconnecting.

Together: Emacs pauses at `ask_user` → open ECA Web later → see the question and options → answer → the original session continues.

No permanent SSE watcher or shared-webview source change is needed. Acknowledgment-first answer UI and a broader cross-client dismissal protocol are outside this change; the existing optimistic answer UI remains unchanged.

## Verification

- `clojure -M:test --focus eca.remote.messenger-test --focus eca.remote.handlers-test --focus eca.features.tools.ask-user-test`: **41 tests, 233 assertions, no failures**, rerun after rebasing onto current upstream.
- Coverage includes questions with no subscribers, immediate editor responses, competing answers, error propagation, and request-handle cleanup.
- A mounted-browser smoke test with both changes verifies late opening, reload with the same request UUID, an option click completing the original waiter exactly once (`204`), and a duplicate answer returning `404` without changing the result.

The browser test used a disposable synthetic editor with the real `ask_user` handler and HTTP/SSE routes, not the active Emacs session or a full LLM loop. Scheduler completion was simulated after the answer. Chrome required `--disable-features=LocalNetworkAccessChecks` for the loopback fixture, so this does not verify default browser networking policy. The full backend suite and native-image build were not run.

## Checklist

- [ ] This is not an AI slop.

🤖 Generated with [ECA](https://eca.dev) (openai/gpt-6-astra - xhigh)


ericdallo (approved):
Good catch!

